PR #1, opened 2023-04-24
Add `login` command

akshatdalton:
This commit adds the `login` command for the
user to log in to the server address passed.

This commit also adds the utils functions that
will be used by successive commits.


Vad1mo:
Functionality wise, I would expect here some similar behavior, like the `docker login` or kubectl where you can work on different clusters and switch between them 

- You can log in to multiple registries
- When you do an operation like a project, create it happens in the context of the current active registry

akshatdalton:
Rather we can give an option to the user to save the config in the path specified by the user and later they can use that config to execute other commands. Reference: https://stackoverflow.com/a/55635346.

Vad1mo:
> Thanks for the review @Vad1mo!
> 
> In kubetctl, if someone is logged in to multiple registries then the arguments are such that it uses creds accordingly, for e.g.:
> 
> if they executed:
> 
> ```
> kubectl create deployment myapp --image=registry.example.com/myimage:1.0
> ```
> 
> it uses creds for `registry.example.com`
> 
> and then if they executed:
> 
> ```
> kubectl create deployment myapp --image=registry2.example.com/myimage:1.0
> ```
> 
> it uses creds for `registry2.example.com`.
> 
> Can you explain the usage in our case? I didn't get it in the create project example you have given.

You have to think more from the perspective of:

`kubectl get pods` that is executed in the context of cluster A or cluster B

In k8s you do it by switching the context via env var or `kubectl config set-context` in the config.  It is similar for docker CLI nowadays. 


The idea you found on https://stackoverflow.com/a/55635346. is an option. However, this is not a very common practice

`export HARBOR_CONFIG=~/.harbor/config_cluster_b.yaml`

From the effort and complexity perspective, you need to ask yourself, should the switch happen via env var or via CLI or maybe both. In K8s both options are possible. 

So to keep the effort reasonable and the solution usable, at least support context switching via env var `HARBOR_CONFIG`if you think switching context like with kubectl is easy to implement and you want to do it go for it.

Issue #11, opened 2024-02-06
Result Presentation in YAML, JSON and Table 


Vad1mo:
Similar to kubectl we want users of the harbor-cli users to provide the same representations and output options.

YAML, JSON, Table.

Currently, only JSON output is supported:

```shell
 ./harbor get project testproject-832165132
{
  "Payload": {
    "chart_count": 0,
    "creation_time": "2024-02-05T14:58:53.974Z",
    "current_user_role_ids": null,
    "cve_allowlist": {
      "creation_time": "0001-01-01T00:00:00.000Z",
      "id": 21,
      "items": [],
      "project_id": 20,
      "update_time": "0001-01-01T00:00:00.000Z"
    },
    "metadata": {
      "public": "true"
    },
    "name": "testproject-832165132",
    "owner_id": 20,
    "owner_name": "koenc",
    "project_id": 20,
    "repo_count": 2,
    "update_time": "2024-02-05T14:58:53.974Z"
  }
}
```

- [ ] Support YAML, JSON and Table output
- [ ] Make Table output default 
- [ ] -o wide: This parameter allows us to display extra information (if needed, otherwise -o wide should have no effect)
- [ ] In doubt, follow kubectl approach 

### Bonus points

- [ ] --sort-by: This parameter allows us to sort the output based on the resource attribute we specify.

### Out of Scope 

- Support of JSONPath, or GO template.
- -f, --field-selector=<attributename>=<attributevalue>: This parameter allows us to filter the output using some resource attributes.

bishal7679:
@Vad1mo Assign this issue to me!
Would like to give it a try

PR #4, opened 2024-01-28
[WIP] feat: revamp harbor cli

amands98:
Enhancements to the CLI:

- [x] Implemented a refined project structure for improved organization.
- [ ] Integrated GitHub release workflows for streamlined version management.
- [ ] Added Docker support 
- [ ] Implemented support for codecov to ensure comprehensive code coverage.
- [ ] Established a robust release cycle management system for the CLI.
- [ ] Support for e2e tests & docs
- [ ] Automation of docs generation 
- [ ] Refining the current code.

PR #17, opened 2024-02-13
Isolate all the cobra cli commands under `cli` Folder

bishal7679:
# Part of fixing #11

## Previous State
- Previously, cobra CLI commands/flags were bit messed up
- Codebase was unstructured

#### part 1
- [x] Restructured the whole codebase with efficient manner
- [x] Separated all the cobra functions eg. `func CreateProjectCommand() *cobra.Command` to `cli/cmd` folder with `root.go` which is for adding all the cobra cli func with `RootCmd`. This cobra func is using all the `RunFunction`  from `api` folder i.e. `apis.go`


JaeAeich:
IMO, the code base structure follows the recommended guidelines, especially using the `internal` folder this PR removes.
Maybe some things can be restructured, but I don't think it is needed. 

PS: I don't want to be intrusive, just an opinion.

bishal7679:
> IMO, the code base structure follows the recommended guidelines, especially using the `internal` folder this PR removes. Maybe some things can be restructured, but I don't think it is needed.
> 
> PS: I don't want to be intrusive, just an opinion.

Actually here the main aim is to separate all the jargons cli, api which would be more efficient for any contributor also to understand the codebase easily. And PR #16 adding `internal/pkg` folder which is also signifying all the internal package being utilized!
WDYT?

amands98 (changes requested):
Let's keep the current folder structure; restructuring isn't necessary as effort has already been invested in this [PR](https://github.com/goharbor/harbor-cli/pull/4).

bishal7679:
> Let's keep the current folder structure; restructuring isn't necessary as effort has already been invested in this [PR](https://github.com/goharbor/harbor-cli/pull/4).

Let's wait for @Vad1mo what this PR seems to him as it also took a lot of effort and interrelated with #16 to separte the whole api and cli. Doesn't it make this codebase more enhanced than previous state in order to understand by any new contributor?

amands98:
closing this PR, since it no longer required. // @Vad1mo

PR #18, opened 2024-02-13
Adding Building script for building Harbor-CLI very easily

bishal7679:
### Fixing
This script will help to build this project and move the binary to executable location without any effort


bishal7679:
> We already have it [here](https://github.com/goharbor/harbor-cli/blob/main/Makefile). What more does this offer?

Hey @amands98 This script is for local building and not any distro specific buuilding

PR #20, opened 2024-02-15
bump: fix the goreleaser release cycle & Dependabot

amands98:
This PR fixes the [failing workflow](https://github.com/goharbor/harbor-cli/actions/runs/7905056008)

This PR suggests tagging `v0.1.0-beta.1` and proceeding with the release.

Updated goreleaser configuration, @Vad1mo . Add **GH_TOKEN** secret with `repo scope, write:packages`.
Also I included a dependabot, to automate dependency updates.

After merging, let's release the packages on Homebrew, Winget, Nix, and Snap.

### DIFF
--- a/.github/workflows/publish_release.yml
+++ b/.github/workflows/publish_release.yml
@@ -11,16 +11,16 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
       - name: Setup go env
         uses: actions/setup-go@master
         with:
-          go-version: "1.20"
+          go-version: "1.21"
       - name: goreleaser with tag
         uses: goreleaser/goreleaser-action@v5
         with:
           version: latest
           args: release --rm-dist
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
-
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
--- a/.github/workflows/publish_release.yml
+++ b/.github/workflows/publish_release.yml
@@ -23,4 +23,4 @@ jobs:
           version: latest
           args: release --rm-dist
         env:
-          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: "/"
+    schedule:
+      interval: daily
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,10 +1,14 @@
 project_name: harbor
 
+before:
+  hooks:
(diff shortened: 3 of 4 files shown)


JaeAeich:
It would also make a lot of sense IMO to make a release script that Unix users can pipe into `bash`/`zsh` to `make` the release form source and configure the binaries into the right directories, something like `~/usr/local/bin` .

bishal7679:
> It would also make a lot of sense IMO to make a install script that Unix users can pipe into `bash`/`zsh` to `make` the release form source and configure the binaries into the right directories, something like `~/usr/local/bin` .

Yes! #18 doing the same thing but need to clone the repository.

Vad1mo:
I need to check that @OrlinVasilev, I don't have the owner permission set for the org to request such a token
.
.

OrlinVasilev:
will take care of that :)

OrlinVasilev:
it should be fine of first place.
image
not sure why are you renaming the token to GH_TOKEN ? what's the idea behind it that comes by default

OrlinVasilev:
@amands98 check this run after the fix https://github.com/goharbor/harbor-cli/actions/runs/7905056008/job/21984205486

.github/workflows/publish_release.yml
```diff
@@ -11,16 +11,16 @@ jobs:
...
         with:
-          go-version: "1.20"
+          go-version: "1.21"
       - name: goreleaser with tag
         uses: goreleaser/goreleaser-action@v5
         with:
           version: latest
           args: release --rm-dist
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
OrlinVasilev, line 24: Keep it as it is ! that new token is not necessary!

OrlinVasilev (changes requested):

bishal7679:
> image not sure why are you renaming the token to GH_TOKEN ? what's the idea behind it that comes by default

It automatically fetches if we specify as GITHUB_TOKEN otherwise for different naming we need to set that as repo secret var.

amands98:
just following the naming rules.
https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#naming-your-secrets

.github/workflows/publish_release.yml
Vad1mo: Does the default GITHUB_TOKEN have  `write:packages` permission? If so we are fine

Vad1mo:
> just following the naming rules. https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#naming-your-secrets
> 
> image

Makes sense if you are creating your own secret, however we are reusing the exising one here.

.github/workflows/publish_release.yml
amands98: `repo scope, write:packages`

Vad1mo:
@amands98 Can you address my comment?

amands98:
Sure, let's use the default `GITHUB_TOKEN` with the required scopes (`repo, write:packages`). After that, we can merge and re-run the workflow to ensure the harbor-cli binaries are generated effectively.

Issue #8, opened 2024-02-06
Empty configuration during `create` panics the binary, but ideally should output actionable data


AbhinavMir:
hi! I was trying to set this up and seems like [credential resolution](https://github.com/goharbor/harbor-cli/blob/ec4cc6b2f723daa7e5dd47d1fd9d68eb7507ed96/cmd/utils/utils.go#L24) panics the binary instead of outputting something like "Empty ~/,harbor/config, please <enter fix>" - the patch for this is rather easy, should I start a PR or is this a candidate for the mentorship program?

Cheers and thanks!
Abhinav.

JaeAeich:
Hey @AbhinavMir and @Vad1mo.
I was just checking in on the issue, wanted to see if this is an OK solution :) (if so let me know I'll create a PR)
Given the error we can have at this point are regarding config issues it would make sense to exit quitely. 

```go
//pkg/utils/credential_store.go

func resolveCredential(credentialName string) (Credential, error) {
	credentialStore, err := readCredentialStore()
	if err != nil {
		return Credential{}, fmt.Errorf("failed to read credential store: %s", err)
	}

...
	if credentialName == "" {
		return Credential{}, fmt.Errorf("current credential name not set, please login again")
	}
...
	return Credential{}, fmt.Errorf("no credential found for the name: %s, please login again with the credential name: %s", credentialName, credentialName)
}

//pkg/utils/utils.go
func GetClientByCredentialName(credentialName string) *v2client.HarborAPI {
	credential, err := resolveCredential(credentialName)
	if err != nil {
		fmt.Print(err) // If error come from resolveCredential, it make sense to exist and print the user friendly message.
		os.Exit(1)
	}
...
}
```

AbhinavMir:
Hi, I was planning to close this myself, but good for you! I am not a reviewer, and have no idea if this is the correct solution. Possibly start a PR for a review, I'm not aware of the review etiquette of this repo.

Cheers.

amands98:
already fixed
// @Vad1mo

PR #21, opened 2024-02-16
fix: empty config panic

JaeAeich:
## Description
While login if the config read fails or the config is not found, it makes more sense to not panic but rather print a user-friendly message and exit out.

This PR also removes the deprecated use of [ioutils](https://github.com/grpc/grpc-go/pull/5906).

Fixes - #8

### DIFF
--- a/pkg/utils/credential_store.go
+++ b/pkg/utils/credential_store.go
@@ -2,7 +2,6 @@ package utils
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


Vad1mo (approved):
awesome!

OrlinVasilev:
@JaeAeich congrats on merging your first PR here :))

Issue #9, opened 2024-02-06
Awkward Harbor login flow: Schema is mandatory.


Vad1mo:
I don't think the schema should be mandatory.

```sh
 ./harbor login demo.goharbor.io -u admin -p Harbor12345
Error: login failed, please check your credentials: Get "https://localhost/demo.goharbor.io/users/current": dial tcp 127.0.0.1:443: connect: connection refused
Usage:

```

- [ ] Use HTTPS by default. (secure by design)
- [ ] Use only HTTP if explicitly specified.
- [ ] Add option to ignore errors in self signed certs. eg. `(--inecure`) [like in curl](https://www.cyberciti.biz/faq/how-to-curl-ignore-ssl-certificate-warnings-command-option/)

bishal7679:
I faced the same issue! Let me see how I can fix it

bupd:
This has been fixed. @Vad1mo, we can close this issue. is there anything else you'd like to add? if there you can reopen this issue.

PR #26, opened 2024-02-24
feat: add insecure flag

JaeAeich:
## Description
Add `--insecure` flag to `login` command.

- Fixes: #9 

Note the issue mentioned an awkward login flow when URL is given without `https` the client defaults to pretending the URL with `localhost` which causes the `cli` to err.

## To consider
Other things to consider opening an issue for:
- Encrypting the password when store in `~/.harbor/config`
- Removing password as flag from `login` command and take it as input discreetly.
- Currently flag `--name` isn't doing much, use to store multiple users and then use this config to easily login for different users, fix `--name` flag iverwritting the previously stored creds.

## Current behaviour
```bash
 ./harbor login https://demo.goharbor.io -n jaeaeich -u admin -p Harbor12345 -i
```

image
image

@Vad1mo If the above mentioned issues seem reasonable maybe its worth opening issues for, do tell me so I can write further details.

### DIFF
--- a/cmd/harbor/root/login.go
+++ b/cmd/harbor/root/login.go
@@ -15,6 +15,7 @@ type loginOptions struct {
 	serverAddress string
 	username      string
 	password      string
+	insecure      bool
 }
 
 // LoginCommand creates a new `harbor login` command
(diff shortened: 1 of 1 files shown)


JaeAeich:
@Vad1mo @OrlinVasilev, can you have a look here, for the --insecure flag, I think this PR does that but maybe I can also push schema mandate issue in this or create another one. Please do let me know also what is the schema that you expected.  I thought that if config is available and there is only one `--name` that should be take as default and login with that, otherwise ask the user for `--name` else if there is no config or it is incomplete ask the user for password and username.

Vad1mo:
I'll look at it in more details in the next couple of days.

PR #13, opened 2024-02-12
added password-stdin flag for login

ZiyanK:
Solved issue: https://github.com/goharbor/harbor-cli/issues/10

PR #33, opened 2024-03-26
feat: improve the login flow for the user

amands98:
it fits perfectly into CI/CD pipelines and works both ways.

image


amands98:
@Vad1mo @OrlinVasilev

Vad1mo:
looks good, but I would also like to have the stdin option too https://github.com/goharbor/harbor-cli/pull/13

amands98:
Ok I will add it.

PR #35, opened 2024-04-01
feat: improve the Project flow in cli

amands98:
- [x] List
image

- [x] Delete

image

- [x] Create

image

- [ ] View

In the view where we display project information, we will also list the scanners and repositories associated with that project.


amands98:
@Vad1mo

Vad1mo (commented):
keep in mind that the harbor CLIs should also be usable in non-interactive mode, In CI/CD pipelines, in scripts. piping in yaml/json and outputting jsons/yaml are equally important.

amands98:
Test Case : 

- [ ] Project
- [ ] Registry

amands98:
We handle projects in this PR only. Will open a separate one for the registry. I've also added subcommands `logs` to the project command recently.

amands98:
> keep in mind that the harbor CLIs should also be usable in non-interactive mode, In CI/CD pipelines, in scripts. piping in yaml/json and outputting jsons/yaml are equally important.

It will works both ways.

amands98:
added table for project logs -
image

PR #51, opened 2024-05-17
handled login server address schema

### LINKED ISSUE #38: Login fails to use schema automatically
Log in into harbor should work without specifying a schema. 

```shell
./harbor login 8gears.container-registry.com
login failed, please check your credentials: Get "https://localhost/8gears.container-registry.com/users/current": dial tcp [::1]:443: connect: connection refused
``` 

The following combinations of commands should all work with schema, without schema, port number.

- [ ] ./harbor login http://8gears.container-registry.com -> http port 80
- [ ] ./harbor login http://8gears.container-registry.com:5000 -> http port 5000
- [ ] ./harbor login https://8gears.container-registry.com -> https port 443
- [ ] ./harbor login 8gears.container-registry.com -> assume it is https port 443
- [ ] ./harbor login 8gears.container-registry.com:8443 -> assume it is https port 8443

muaz-32:
Signed-off-by: muaz-32 <96006730+muaz-32@users.noreply.github.com>

Fixes #38 
@Vad1mo please give it a check

### DIFF
--- a/cmd/harbor/root/login.go
+++ b/cmd/harbor/root/login.go
@@ -3,12 +3,13 @@ package root
 import (
 	"context"
 	"fmt"
-
 	"github.com/goharbor/go-client/pkg/harbor"
 	"github.com/goharbor/go-client/pkg/sdk/v2.0/client/user"
 	"github.com/goharbor/harbor-cli/pkg/utils"
 	"github.com/goharbor/harbor-cli/pkg/views/login"
 	"github.com/spf13/cobra"
+	"net/url"
+	"strings"
 )
 
 var (
@@ -81,8 +82,12 @@ func createLoginView(loginView *login.LoginView) error {
 }
 
 func runLogin(opts login.LoginView) error {
+	server, URLFormatErr := handleURLFormat(opts.Server)
+	if URLFormatErr != nil {
+		return URLFormatErr
+	}
 	clientConfig := &harbor.ClientSetConfig{
-		URL:      opts.Server,
+		URL:      server,
 		Username: opts.Username,
 		Password: opts.Password,
 	}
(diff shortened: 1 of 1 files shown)


cmd/harbor/root/login.go
```diff
@@ -106,3 +111,22 @@ func runLogin(opts login.LoginView) error {
 	}
 	return nil
 }
+
+func handleURLFormat(server string) (string, error) {
```
amands98, line 115: Can we put it down to helper function.

muaz-32: Done.

bishal7679:
This is still throwing some error
- it's nice that `https` is getting attached as default schema if it's missing, but it's not being added to the `config.yaml` file in place of `serveraddress` key.
So when we are calling the api for listing/creating or any operation for any resources it is taking localhost as prefix in the url everytime and hence causing the error



## Solution

After a lot of debugging finally I've found that you are not updating the serveraddress here in runLogin function which will update the `serveraddress` key in `config.yaml` file via `AddCredentialsToConfigFile` function



I hope it helps :)

muaz-32:
> This is still throwing some error
> 
> * it's nice that `https` is getting attached as default schema if it's missing, but it's not being added to the `config.yaml` file in place of `serveraddress` key.
>   So when we are calling the api for listing/creating or any operation for any resources it is taking localhost as prefix in the url everytime and hence causing the error
> 
> 
> 
> ## Solution
> After a lot of debugging finally I've found that you are not updating the serveraddress here in runLogin function which will update the `serveraddress` key in `config.yaml` file via `AddCredentialsToConfigFile` function
> 
> 
> 
> I hope it helps :)

Thanks, updated the pull.

bupd (commented):
I have tested these changes in my local environment, and everything works as expected :+1:

amands98 (approved):
:rocket:

Vad1mo (changes requested):

PR #61, opened 2024-05-19
feat: Added test cases for the `project` package.

Mehul-Kumar-27:
This PR adds test cases for the cmd/harbor/root/project package.

Note: Test cases for the `create` command are excluded due to a bug (see [PR#60](https://github.com/goharbor/harbor-cli/pull/60)), as the flow might change.


muaz-32 (commented):
Your approach to e2e testing involves mocking the functions. The purpose of e2e test is to check if the whole process is working properly, from one end to the other end. Using mocks does not allow to check the integration of all the modules. You can have a look at this [PR](https://github.com/goharbor/harbor-cli/pull/90)

Issue #63, opened 2024-05-20
Fixing user create command issue


Althaf66:
When we create a new user, there is some guidelines for accepting the password like
- password should be 8 character long
- need atleast one uppercase
- need atleast one lowercase
- need atleast one number

If the user violate one of the guideline, we cannot create a new user.
![Screenshot from 2024-05-21 02-28-25](https://github.com/goharbor/harbor-cli/assets/114910365/c85771fb-d13a-4e80-b00a-5fe5ad88832c)

**Expected Result:**
Required proper error message when user violate these guidelines at the time we enter the password.

PR #64, opened 2024-05-20
enhancement:user create command

Althaf66:
This issue fixes the #63 
 @Vad1mo @amands98 
This PR give a proper error message when user forgot to add one lowercase,uppercase,number in their password.


pkg/views/user/create/view.go
```diff
@@ -54,8 +56,17 @@ func CreateUserView(createView *CreateView) {
 				Validate(func(str string) error {
 					if str == "" {
 						return errors.New("password cannot be empty")
-					}
+					} else if len(str) < 8  {
```
amands98, line 59: Thank you for your effort here.

We can just use regexp here, no need for a custom validator. Apply regexp to check the password and show a simple error message like 'Password is incorrect'. Also, let's add a note in the input field description about the expected password format.

pkg/views/user/create/view.go
```diff
@@ -67,3 +72,30 @@ func CreateUserView(createView *CreateView) {
 		log.Fatal(err)
 	}
 }
+
+
+func IsValidPassword(password string) bool {
+	lengthPattern := `^.{8,128}$`
```
amands98, line 78: Can you merge the regex into one?

Althaf66: Combining the regex causing error. but, I optimized the PR better!

pkg/views/user/create/view.go
```diff
@@ -67,3 +72,18 @@ func CreateUserView(createView *CreateView) {
 		log.Fatal(err)
 	}
 }
+
+
+func IsValidPassword(password string) bool {
+	if !regexp.MustCompile(`^.{8,128}$`).MatchString(password) {
```
amands98, line 78: This can be enhanced by employing a single regex.

Althaf66: Go supports RE2 syntax, so lookaround assertions is not support in RE2. We can use another packages for checking validation.
[https://support.google.com/docs/thread/215946555/data-validation-for-password-must-have-at-least-one-upper-case-character?hl=en](https://support.google.com/docs/thread/215946555/data-validation-for-password-must-have-at-least-one-upper-case-character?hl=en)

bupd: https://github.com/goharbor/harbor-cli/blob/a0a9b98b8a5e04c7560f1a6710b437679f42d8cd/pkg/utils/utils.go#L62

You can use the above method instead of relying on regexp.

### Since, considering performance of both methods:
- **Character-by-Character Validation:** This method might be slightly more **performant** because it only iterates through the string once and uses direct character checks.

- **Regular Expressions:** This method involves multiple regex compilations and matches. Each MatchString call processes the string separately, which can be less efficient.

> Note: go regexp doesn't support backtracking.

pkg/views/user/create/view.go
```diff
@@ -54,9 +55,13 @@ func CreateUserView(createView *CreateView) {
 				Validate(func(str string) error {
 					if str == "" {
 						return errors.New("password cannot be empty")
-					}
+					} else if !IsValidPassword(str)  {
+					 	return errors.New("password is incorrect")
```
bupd, line 59: improve the error message on user create.

`Password is incorrect` is misleading.

`Password should contain 8-128 characters long with at least 1 uppercase, 1 lowercase and 1 number.` would be a good error message or

bupd (changes requested):
Improve error message.

bupd:
@Althaf66 would be good if we can add the validation for email too.

bupd:
This PR can be closed since the input fields are validated by #198

Issue #65, opened 2024-05-21
Bug: newline added for inputs while login in bash


muaz-32:
When using bash in Windows, a new line is added with the input when enter is pressed for the next input. As a result, an extra space character is added to the server URL. It happens for all the inputs probably. For example, the password is changed due to carriage return so the user is unable to log in. Without pressing enter, the user cannot move to the next input. So it would be better to trim the spaces of the inputs.

System: Windows 10
Terminal: Bash

amands98:
Could you record the video or Screen shot and explain what the problem was?

muaz-32:
> Could you record the video or Screen shot and explain what the problem was?


https://github.com/goharbor/harbor-cli/assets/96006730/218b4690-5629-47d0-be60-8b9656d92e32

In the video, you can see that in the beginning an extra space is added with the server URL. So when I trimmed the spaces, it was an authentication error because the password was carrying one extra character (carriage return).

I have updated the PR description.

bupd:
This has been fixed.

PR #66, opened 2024-05-21
Fixed the extra space issue in inputs when logging in.

muaz-32:
Fixes: #65 

Could not place the Trimming function in helper.go file as it creates cyclic dependency.
@amands98 Please give it a check.

### DIFF
--- a/pkg/utils/helper.go
+++ b/pkg/utils/helper.go
@@ -2,6 +2,8 @@ package utils
 
 import (
 	"fmt"
+	"github.com/goharbor/harbor-cli/pkg/views/login"
+	"strings"
 	"time"
 )
 
@@ -24,4 +26,11 @@ func FormatCreatedTime(timestamp string) (string, error) {
 	} else {
 		return fmt.Sprintf("%d day ago", days), nil
 	}
+}
+
+func TrimSpaceOfLoginInputs(loginView *login.LoginView) {
+	loginView.Server = strings.TrimSpace(loginView.Server)
+	loginView.Username = strings.TrimSpace(loginView.Username)
+	loginView.Password = strings.TrimSpace(loginView.Password)
+	loginView.Name = strings.TrimSpace(loginView.Name)
 }
--- a/pkg/views/login/create.go
+++ b/pkg/views/login/create.go
@@ -2,7 +2,6 @@ package login
 
 import (
 	"errors"
-	"fmt"
 	"strings"
 
 	"github.com/charmbracelet/huh"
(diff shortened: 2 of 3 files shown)


muaz-32:
@amands98 I have updated the commit. Removed the line that prints the credentials in the console.

Vad1mo:
Thank you, this makes sense 
One question: Does huh not have an option to trim on the input field?

pkg/views/login/create.go
```diff
@@ -61,5 +62,12 @@ func CreateView(loginView *LoginView) {
 	if err != nil {
 		log.Fatal(err)
 	}
+	TrimSpaceOfLoginInputs(loginView)
+}
 
+func TrimSpaceOfLoginInputs(loginView *LoginView) {
```
amands98, line 68: Transfer this logic to a helper function and invoke it directly within the `login.go` file

muaz-32: Done.

muaz-32:
> Thank you, this makes sense One question: Does huh not have an option to trim on the input field?

No. Afaik huh does not provide any such functionality.

Vad1mo:
IMO, this should be part of the validation.  Try testing with "   ". As this is part of the input validation process or pre validation.

muaz-32:
> IMO, this should be part of the validation. Try testing with " ". As this is part of the input validation process or pre validation.

In that case, `strings.TrimSpace()` can be directly called in the CreateView function after taking the inputs, without creating another function to trim all the inputs. Should I do it that way?

Also, If I call the `strings.TrimSpace()` function inside the huh validate method, it does not work as the validate method does not return the modified string. After much digging, I came to conclude that huh currently does not support it.

muaz-32:
@amands98 I have resolved the conflict with the newer changes.

pkg/utils/helper.go
```diff
@@ -25,3 +27,10 @@ func FormatCreatedTime(timestamp string) (string, error) {
 		return fmt.Sprintf("%d day ago", days), nil
 	}
 }
+
+func TrimSpaceOfLoginInputs(loginView *login.LoginView) {
+	loginView.Server = strings.TrimSpace(loginView.Server)
+	loginView.Username = strings.TrimSpace(loginView.Username)
+	loginView.Password = strings.TrimSpace(loginView.Password)
+	loginView.Name = strings.TrimSpace(loginView.Name)
+}
```
bupd, line 36:
```suggestion

```


we don't need separate function to trim space of inputs. Since we can use `huh.Validate()` will work fine.

bupd (changes requested):
```go
huh.NewInput().
	Title("Server").
	Value(&loginView.Server).
		trimmedStr := strings.TrimSpace(str)
		loginView.Server = trimmedStr
		if trimmedStr == "" {
			return errors.New("server cannot be empty or only spaces")
		}
		return nil
	}),
```

## Changes to be done:
- Remove redundant function `TrimSpaces...`
- Update `huh.Validate()` as mentioned above.

bupd:
This PR can be closed since the input fields are validated by #198

PR #73, opened 2024-05-24
feat: Fix Creating Project Flow

### LINKED ISSUE #12: Improve Creating Project flow.
When creating a project as a user, I would expect that there is no need to explicitly provide the --name  parameter.

 So something like this ./harbor create project vadim-test should work.
 --public should be by default `false``
 The description and logic behind --registry-id int is not obvious/clear. When id !=1 does it mean that than a proxy project is created?
./harbor create project vadim-test
Error: [POST /projects][500] createProjectInternalServerError  &{Errors:[0x14000130020]}
Usage:
  harbor create project [flags]

Flags:
  -h, --help                help for project
      --name string         Name of the project
      --public              Project is public or private (default true)
      --registry-id int     ID of referenced registry when creating the proxy cache project (default 1)
      --storage-limit int   Storage quota of the project (default -1)

bupd:
### Description:
- Allow project name to be specified as a positional argument in the `project create` command.
- Change the default value of the `--public` flag to `false`, making projects private by default.
- Remove the `--name` flag and update the command to handle the project name from the positional argument.

### Examples:
- Create a private project by default:
  `./harbor project create bupd-test` will work

- Create a public project:
  `./harbor create project bupd-test --public` 

These changes improve user experience by simplifying command usage and enhance security by setting a private default for projects.

### DIFF
--- a/cmd/harbor/root/project/create.go
+++ b/cmd/harbor/root/project/create.go
@@ -12,9 +12,11 @@ func CreateProjectCommand() *cobra.Command {
 	var opts create.CreateView
 
 	cmd := &cobra.Command{
-		Use:   "create",
+		Use:   "create [project name]",
 		Short: "create project",
+		Args:  cobra.ExactArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
+			opts.ProjectName = args[0]
 			var err error
 			createView := &create.CreateView{
 				ProjectName:  opts.ProjectName,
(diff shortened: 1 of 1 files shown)


bupd:
ready for review @Vad1mo  @amands98

cmd/harbor/root/project/create.go
```diff
@@ -18,10 +18,11 @@ func CreateProjectCommand() *cobra.Command {
 	var opts create.CreateView
 
 	cmd := &cobra.Command{
-		Use:   "create",
+		Use:   "create [project name]",
 		Short: "create project",
-		Args:  cobra.NoArgs,
+    Args:  cobra.ExactArgs(1),
```
amands98, line 23:
```suggestion
    Args:  cobra.NoArgs,
```

cmd/harbor/root/project/create.go
```diff
@@ -18,10 +18,11 @@ func CreateProjectCommand() *cobra.Command {
...
 		Run: func(cmd *cobra.Command, args []string) {
+			opts.ProjectName = args[0]
```
amands98, line 25:
```suggestion
```

cmd/harbor/root/project/create.go
```diff
@@ -45,8 +46,7 @@ func CreateProjectCommand() *cobra.Command {
 	}
 
 	flags := cmd.Flags()
```
amands98, line 48:
```suggestion
	flags := cmd.Flags()
	flags.StringVarP(&opts.ProjectName, "name", "", "", "Name of the project")

```

amands98:
heads up : merge conflicts

cmd/harbor/root/project/create.go
bupd: What's the tradeoff over using flags instead of args?

amands98: to have a consistent behaviour across other resources like registry.

PR #74, opened 2024-05-25
Added presentation formats (JSON and YAML) with some small modifications.

muaz-32:
Fixes part of: #11 

It seems not all commands should have this option. In Kubectl, the `get` command has this output format option. `get` returns the resources, mainly it returns some information. In Kubectl the option is not set as a global option. It is added where it is necessary or can have this feature. So in harbor CLI, the option should be available for some specific commands. In this PR I have added it for the `list` command. In the current code, the `-o` option is given as a global option. So I think it should be moved to the specific commands. So I moved the option to the `list` commands.

The default return format of the lists from the go client is table, so the output format `table` is not added as an option. It is the default return format for the list command. The user can only specify JSON or YAML.

Refactored the `list.go` files of the commands to make them similar and symmetric. 

Renamed the `ListRegistry` to `ListRegistries`, converted to plural to make things symmetric.

### DIFF
--- a/cmd/harbor/root/cmd.go
+++ b/cmd/harbor/root/cmd.go
@@ -16,7 +16,6 @@ import (
 )
 
 var (
-	output  string
 	cfgFile string
 	verbose bool
 )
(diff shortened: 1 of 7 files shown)


muaz-32:
@amands98 @Vad1mo I have updated the PR. It is more suitable and user-friendly to provide the project name as an argument than as an option in `repo list` command. So the user will provide the project name as an argument.

muaz-32:
@amands98 please give it a check. I have updated the PR based on newer commits.

PR #86, opened 2024-05-28
Enhance `registry create` Command

bupd:
## Overview:
This PR improves the user experience of the registry create command by fetching and presenting default registries available. This enhancement guides users to select the correct registry provider, reducing the likelihood of errors and streamlining the setup process.

### Problem: 
Previously, users had to manually enter registry provider details, which often led to errors and a frustrating experience. Incorrect registry configurations could result in errors that were difficult to diagnose and fix.

### Solution:
To address this issue, I have implemented a feature that fetches a list of registry providers in harbor and presents them to the user during the registry create process. This allows users to select a registry from a predefined list, ensuring correct configuration and reducing errors.

### Changes Made:
- Added functionality to fetch a list of default registries and an interactive list of registries for the user to choose from.

### Before:
Before

### After:
after

### DIFF
--- a/pkg/views/registry/create/view.go
+++ b/pkg/views/registry/create/view.go
@@ -2,39 +2,59 @@ package create
 
 import (
 	"errors"
+	"fmt"
+	"strconv"
 
 	"github.com/charmbracelet/huh"
+	"github.com/goharbor/harbor-cli/pkg/api"
 	log "github.com/sirupsen/logrus"
 )
 
-type CreateView struct {
-	Name        string
-	Type        string
-	Description string
-	URL         string
-	Credential  RegistryCredential
-	Insecure    bool
+// struct to hold registry options
+type RegistryOption struct {
+	ID   string
+	Name string
 }
 
-type RegistryCredential struct {
-	AccessKey    string `json:"access_key,omitempty"`
-	Type         string `json:"type,omitempty"`
-	AccessSecret string `json:"access_secret,omitempty"`
-}
+func CreateRegistryView(createView *api.CreateRegView) {
+	registries, _ := api.GetRegistryProviders()
 
-func CreateRegistryView(createView *CreateView) {
+	// Initialize a slice to hold registry options
+	var registryOptions []RegistryOption
+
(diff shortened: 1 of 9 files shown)


bupd:
## Updated:
- Allows registry provider selection.

afterReg

bupd:
looks good to merge @Vad1mo

pkg/views/registry/create/view.go
```diff
@@ -22,19 +28,67 @@ type RegistryCredential struct {
...
 }
 
+// Define a struct to hold registry options
+type RegistryOption struct {
+	ID   string
+	Name string
+}
+
+func getRegistryProviders() ([]string, error) {
+	credentialName := viper.GetString("current-credential-name")
```
amands98, line 38: use this method `ContextWithClient`

amands98: add this to registry_handler.go

bupd:
Resolved requested Changes.

amands98:
please run gofmt to lint the code.

pkg/views/registry/list/view.go
```diff
@@ -37,26 +37,6 @@ func ListRegistry(registry []*models.Registry) {
 		})
 	}
 
-	// t := table.New(
```
amands98, line 40: Please donot remove this code. We may require it sometime

bupd: keeping commented-out code in the codebase can lead to confusion and clutter over time.

Since we use Git for version control, any code that we remove can always be retrieved from the repository history if needed. This allows us to maintain a clean and readable codebase while still having access to previous versions of the code if we ever need to reference it.

bupd: If there's a specific reason why you think we might need this code in the future, let's discuss it

amands98: Please delete it; we don’t need it. My apologies—I forgot that I also created a utility for the table list.
https://github.com/goharbor/harbor-cli/blob/main/pkg/views/base/tablelist/model.go#L14

PR #88, opened 2024-05-28
feat: modified registy update command

### LINKED ISSUE #87: fix: input field are empty for update registry
**Problem**

When updating the registry, all the input field are empty.
If the user gets the value in the input field like in web UI, It is more easier to make changes to them.

**Solution**

Values need to shown while using update registry command

Althaf66:
This fixes #87 

Registry values where shown in the input field so, user have easier to update the values in the registry. This fuctionality present in the Web UI


cmd/harbor/root/registry/update.go
```diff
@@ -7,82 +7,67 @@ import (
...
-
+			update.UpdateRegistryView(updateView)
+			err = runUpdateRegistry(updateView, registryId)
 			if err != nil {
 				log.Errorf("failed to update registry: %v", err)
 			}
 		},
 	}
 
-	flags := cmd.Flags()
```
bupd, line 63: Thanks for your contribution! Could you please add back the below flags? They're crucial for  automation and scripting processes, and their absence could potentially disrupt CI/CD pipelines.

Thanks.

bupd (changes requested):

cmd/harbor/root/registry/update.go
Althaf66: Yes. I will add those flags.

amands98:
heads up: merge conflicts

bupd:
@Althaf66 merge conflicts.

otherwise, LGTM 🙌

Vad1mo:
update with empty id and secret shoould work

Vad1mo (approved):

Issue #91, opened 2024-05-28
Add the metadata command for managing project metadata


muaz-32:
The metadata command might be inside the project command (such as `project metadata` rather than `metadata [project]`, as metadata is only for projects and it is more intuitive to access.

PR #92, opened 2024-05-28
Added the metadata command inside project command.

muaz-32:
Fixes: #91 

The metadata will be added directly using the command, without using the interactive cmd. This is the fastest way to add metadata, so it has a better user experience.
For example: `project metadata add [NAME|ID] ...[KEY]:[VALUE]`
The same idea is used for other metadata commands.

flag: `--id` which means whether the argument is project ID or not. It is beneficial when a project's name is the same as another project's ID.


muaz-32:
> Thank you ! Please check out the latest changes we've made, they significantly improve how the handler is used

I have updated it accordingly. Also, I am updating my other PRs.

muaz-32:
> > Thank you ! Please check out the latest changes we've made, they significantly improve how the handler is used
> 
> I have updated it accordingly. Also, I am updating my other PRs.

muaz-32:
> Thank you ! Please check out the latest changes we've made, they significantly improve how the handler is used

Regarding the new changes, how should the flags be handled? For example this PR has the `isBool` flag, but most other commands do not have it. So the `isBool` flag is not included in the `ListFlags` struct. I think we can create individual structs for the commands. WDYT?

Vad1mo:
Metadata is essentially the project specific configuration, as it is marked like that on the UI we should follow the same convention, and name it Configuration.

Example:
```sh
harbor-cli project config add project_name key value
```

bupd (changes requested):

bupd:
closing in favor of https://github.com/goharbor/harbor-cli/pull/448

PR #109, opened 2024-06-06
Added `scanner` command.

muaz-32:
This PR will add all the scanner commands.

### DIFF
--- a/pkg/api/scanner_handler.go
+++ b/pkg/api/scanner_handler.go
@@ -1,0 +1,38 @@
+package api
+
+import (
+	"github.com/go-openapi/strfmt"
+	"github.com/goharbor/go-client/pkg/sdk/v2.0/client/scanner"
+	"github.com/goharbor/go-client/pkg/sdk/v2.0/models"
+	"github.com/goharbor/harbor-cli/pkg/utils"
+	"github.com/goharbor/harbor-cli/pkg/views/scanner/create"
+)
+
+func CreateScanner(opts create.CreateView) error {
+	ctx, client, err := utils.ContextWithClient()
+	if err != nil {
+		return err
+	}
+
+	var scannerRegReq models.ScannerRegistrationReq
+	scannerRegReq.Name = &opts.Name
+	scannerRegReq.Description = opts.Description
+	scannerRegReq.Auth = opts.Auth
+	scannerRegReq.AccessCredential = opts.AccessCredential
+	url := strfmt.URI(opts.URL)
+	scannerRegReq.URL = &url
+	scannerRegReq.Disabled = &opts.Disabled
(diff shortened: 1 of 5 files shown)


pkg/api/scanner_handler.go
```diff
@@ -0,0 +1,50 @@
...
+		return err
+	}
+
+	response, err := client.Scanner.CreateScanner(ctx, &scanner.CreateScannerParams{Registration: &scannerRegReq})
+
+	if err != nil {
+		return err
+	}
+
+	if response != nil {
```
amands98, line 46: Do we need this check if the one above is enough?

pkg/api/scanner_handler.go
```diff
@@ -0,0 +1,50 @@
...
+	url := strfmt.URI(opts.URL)
+	scannerRegReq.URL = &url
+	scannerRegReq.Disabled = &opts.Disabled
+	scannerRegReq.SkipCertVerify = &opts.SkipCertVerify
+	scannerRegReq.UseInternalAddr = &opts.UseInternalAddr
+
+	pingScannerReq.Name = &opts.Name
+	pingScannerReq.Auth = opts.Auth
+	pingScannerReq.AccessCredential = opts.AccessCredential
+	pingScannerReq.URL = &url
```
amands98, line 33:
```suggestion
	scannerRegReq := models.ScannerRegistrationReq{
    Name:             &opts.Name,
    Description:      opts.Description,
    Auth:             opts.Auth,
    AccessCredential: opts.AccessCredential,
    URL:              strfmt.URI(opts.URL),
    Disabled:         &opts.Disabled,
    SkipCertVerify:   &opts.SkipCertVerify,
    UseInternalAddr:  &opts.UseInternalAddr,
}

pingScannerReq := models.ScannerRegistrationSettings{
    Name:             &opts.Name,
    Auth:             opts.Auth,
    AccessCredential: opts.AccessCredential,
    URL:              strfmt.URI(opts.URL),
}
```

amands98: much cleaner

muaz-32: Thank you.

pkg/api/scanner_handler.go
muaz-32: The first check is for pinging the scanner, to see if the URL and auth work correctly. If it does then we create the scanner. 
The second check is for whether there was an error while creating the scanner.

I tried to mimic the interaction in the web UI, as the user will test the connection before creating the scanner. But it seems that the `createScanner` API already tests the connection before creating it.

So yes we can remove the previous ping check.
I am updating the PR accordingly. Thank you for identifying it.

amands98:
please stash the commit .

PR #129, opened 2024-07-16
Add: created immutable command

Althaf66:
Created new command immutable to the Harbor CLI

`harbor immutable create ` - for creating new tag immutability rule 
`harbor immutable list` - list all immutable rule
`harbor immutable delete` - delete specific immutable rule


bupd (changes requested):
```
harbor immutable create
``` 
it is not clear. and will create confusion.

I am thinking of better way to have this immutability

since tag has immutablity and retention.

Why don't we add immutable as subcommand of tag. I am just brainstorming ideas here.

```bash
harbor tag immutable ...
and 
harbor tag retention ...
``` 

Let me know your thoughts on this @Althaf66 @Vad1mo

rizul2108:
@bupd I am continuing work on this PR

Issue #150, opened 2024-08-04
change harbor to harbor-cli in README.md


EraKin575:
during `go build .` command, it creates a binary named `harbor-cli` nor `harbor`. We should change the name to harbor-cli

PR #151, opened 2024-08-04
changed to harbor-cli

EraKin575:
during `go build .` command, it creates a binary named `harbor-cli` nor harbor. We should change the name to harbor-cli. Changed name to `harbor-cli`
Fixes : #150


EraKin575:
yeah, you are right.

Sweetdevil144:
Can you close this PR too for now?

PR #159, opened 2024-08-08
feat: add support for the generation of cli docs for harbor

karaaaanSingh:
Hello Team, 

By adding support for generating YAML CLI documentation for Harbor, we can manage the documentation in one place and easily reference it on the Harbor website.


bupd:
We already have both `man` pages and `markdown` documentation. Could you clarify if there's a specific use case you're thinking of where YAML documentation would be particularly beneficial?

PR #162, opened 2024-08-12
Added detailed instructions for building from source on Windows

has9sayed:
Update Windows build instructions in "build from source"

- Added steps for moving `harbor.exe` to a preferred directory (e.g., `C:\Program Files\harbor\`).
- Included detailed instructions for adding the directory to the PATH environment variable on Windows.


amands98 (commented):
We already have the Windows installation method mentioned below; this seems redundant

PR #93, opened 2024-05-29
Add: Statistics Command

bupd:
## Overview:
This PR adds `stats` command which shows the overall statistic information about the projects and repositories.

PR #176, opened 2024-08-19
feat: add systeminfo cli command

mahbub570:
Commands to interact with the Harbor system information, including general info, volumes, and certificates.

`harbor systeminfo volumes`
image

`harbor systeminfo info`
image

`harbor systeminfo cert`
image

The `systeminfo.GetCertNotFound` error is triggered because the function client.`Systeminfo.GetCert` does not find the certificate it's supposed to retrieve as the certificate endpoint is not available.


Althaf66:
There is another PR #93 with the same command. I think that's better.

PR #180, opened 2024-08-21
chore: add license header to harbor cli project

amands98:
Add license header to harbor cli just like we have for harbor project.


bupd:
then, We can add a workflow to check for headers in every file.

Vad1mo:
we need a more consistant approch to make sure headers are set or not set thema at all.

bupd:
```yaml
run:
  timeout: 3m
linters-settings:
  gofmt:
    # Simplify code: gofmt with `-s` option.
    # Default: true
    simplify: false
  misspell:
    locale: US,UK
  goimports:
    local-prefixes: github.com/goharbor/harbor
  stylecheck:
    checks: [
      "ST1019",  # Importing the same package multiple times.
    ]
  goheader:
    template-path: copyright.tmpl

linters:
  enable:
    # Default linters are already enabled, these are the additional ones
    - gosimple
    - typecheck
    - gofmt
    - goimports
    - gosec
    - nilnil
    - unused
    - errcheck
    - staticcheck
    - dupl
    # - wrapcheck
    # - gocritic
    # - revive #, enable once current issues are resolved
issues:
  exclude-dirs:
    - dagger/internal
  exclude-files:
    - ^.*\\.gen\\.go$
```

with the above we can configure the linter to check if the header is present.

PR #202, opened 2024-10-10
Bump golangci-lint to latest

tylerauerbeck:
Bump golangci-lint to latest release

### DIFF
--- a/dagger/main.go
+++ b/dagger/main.go
@@ -2,18 +2,18 @@ package main
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


tylerauerbeck:
Looks like this publish step has been failing because of missing credentials? Not sure if that's because they don't exist OR they're just not available to PR's from forks.

tylerauerbeck:
`docker-publish` failure should be addressed by: https://github.com/goharbor/harbor-cli/pull/204

Issue #197, opened 2024-09-21
Fix Github Actions


bupd:
Github Actions fail on every PRs with the below errors.
https://github.com/goharbor/harbor-cli/actions/runs/10971891257/job/30467526514?pr=97

## Requirements:

1. Docker Publish Workflow: Update the [docker_publish_workflow](https://github.com/goharbor/harbor-cli/blob/main/.github/workflows/docker_publish.yml) to trigger only on tag creation or release events, preventing runs on PRs.
   
2. Remove Module Dependency: Eliminate the dependency on the module, as seen in [this section](https://github.com/goharbor/harbor-cli/blob/62c3945a773d405a4cc8bfa9bad11031b69c54dc/.github/workflows/pull-request.yml#L42C10-L42C58). It's an unnecessary overhead.

3. Add Build Check Workflow: To check if building binaries for multiple operating systems and architectures are working as expected.

4. Add Formatting Check on Lint Workflow: To maintain consistent formatting.

5. Dagger Flag Update: Replace `--directory-arg` with the `--source` flag in the workflow, and set it as the default using the latest Dagger version.

tylerauerbeck:
@bupd I'll be interested to see if this is still an issue with #201 merged.

bupd:
## To close this issue:

1. Docker Publish Workflow: Update the [docker_publish_workflow](https://github.com/goharbor/harbor-cli/blob/main/.github/workflows/docker_publish.yml) to trigger only on tag creation or release events, preventing runs on PRs.
   
2. Remove Module Dependency: Eliminate the dependency on the module, as seen in [this section](https://github.com/goharbor/harbor-cli/blob/62c3945a773d405a4cc8bfa9bad11031b69c54dc/.github/workflows/pull-request.yml#L42C10-L42C58). It's an unnecessary overhead.

3. Add Build Check Workflow: To check if building binaries for multiple operating systems and architectures are working as expected.

4. Add Formatting Check on Lint Workflow: To maintain consistent formatting.

5. Dagger Flag Update: Replace `--directory-arg` with the `--source` flag in the workflow, and set it as the default using the latest Dagger version.

While #201 addresses some issues and removes unneccessary autogenerated code, it doesn’t cover the above key points which are essential for completely closing this issue as completed.

tylerauerbeck:
Alright, 1, 2 and 5 have been resolved and merged. Will take a look at 3 and 4 next.

PR #206, opened 2024-10-15
use local dagger module in pull-request workflow

tylerauerbeck:
Per #197:

> Remove Module Dependency: Eliminate the dependency on the module, as seen in [this section](https://github.com/goharbor/harbor-cli/blob/62c3945a773d405a4cc8bfa9bad11031b69c54dc/.github/workflows/pull-request.yml#L42C10-L42C58). It's an unnecessary overhead.

> Dagger Flag Update: Replace --directory-arg with the --source flag in the workflow, and set it as the default using the latest Dagger version.

- Drops the usage of a remote `dagger` module and instead looks to use the one local to the repository
- Updates dagger function(s) to use common `source` parameter

### DIFF
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -21,5 +21,4 @@ jobs:
         with:
           version: "latest"
           verb: call
-          module: github.com/bishal7679/harbor-cli@v0.6.4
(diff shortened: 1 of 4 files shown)


tylerauerbeck:
`docker-publish` failure will get solved by #204

PR #211, opened 2024-10-17
add ping command

tylerauerbeck:
Adds the ping command to check connectivity to the Harbor server

### DIFF
--- a/cmd/harbor/root/ping.go
+++ b/cmd/harbor/root/ping.go
@@ -1,0 +1,41 @@
+package root
+
+import (
+	"fmt"
+
+	"github.com/goharbor/go-client/pkg/sdk/v2.0/client/ping"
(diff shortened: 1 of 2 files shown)


cmd/harbor/root/ping.go
```diff
@@ -0,0 +1,41 @@
...
+
+	resp, err := client.Ping.GetPing(ctx, &ping.GetPingParams{})
+	if err != nil {
+		logrus.Errorf("failed to ping the server: %v", err)
+		return err
+	}
+
+	fmt.Printf("Ping: %s\n", resp.Payload)
+	return nil
+}
```
bupd, line 41: I suggest moving this to `pkg/api/ping_handler.go`. to maintain consistency with other commands.

bupd (changes requested):

PR #221, opened 2024-10-25
Dagger Code Review

levlaz:
Hello! 

I went through the existing Dagger module and refactored things a bit to encourage best practices around types, documentation, constructors, etc. 

Here's an overview of the changes. 

1. Refactor GHA workflow to use secrets from env instead of strings 
2. Refactor GHA workflow to not set up go since it is never used directly 
3. Use constructor pattern to clean up functions 
4. More consistently apply code style 
5. Use secrets instead of strings for secrets because this prevents them from ever leaking in logs or cache 
6. Temporarily make build function private due to this bug: https://github.com/dagger/dagger/issues/8202#issuecomment-2317291483 
7. Have the new Build Dev function return a file instead of directory since its redundant.

### DIFF
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -32,4 +32,4 @@ jobs:
         with:
           version: "latest"
           verb: call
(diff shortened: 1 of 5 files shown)


levlaz:
@Vad1mo the only question I have is should we parameterize these constants into the constructor? 

```
const (
	GOLANGCILINT_VERSION = "v1.61.0"
	GO_VERSION           = "1.22.5"
	SYFT_VERSION         = "v1.9.0"
	GORELEASER_VERSION   = "v2.3.2"
)
```

This makes the code a bit more complicated, but also allows us to do things like build-matricies around go versions and overall just make things more testable.

levlaz:
@Vad1mo one more thing, I noticed in the README you have this example but the function does not exist. Is this an oversight?

```
dagger call build-dev --platform darwin/arm64 export --path=.
```

PR #238, opened 2024-11-01
Added flags for user list cmd

Althaf66:
**Added flags for the `user list` command**

Before there were no flags for the user list command, so only a limited number of user lists would be shown.
But now we can view all of the users list using flags


Althaf66:
ready for review

bupd:
Found `Artifact List` also doesn't have flags. would you mind adding it there too.

Vad1mo (approved):

Issue #234, opened 2024-11-01
The test username does not meet the username requirements


JianMinTang:
the user requirements:
<img width="510" alt="image" src="https://github.com/user-attachments/assets/d9de165b-c54d-45d4-88a1-b900a4398ffa">
the ValidateUserName funtion 
<img width="434" alt="image" src="https://github.com/user-attachments/assets/55599dc3-a7f5-4d4c-96ba-9ccf259b21e8">
the username for testing:
<img width="603" alt="image" src="https://github.com/user-attachments/assets/958e745c-7381-47a7-a654-60b1ee66cdda">
We need to standardize these validation requirements, eg `harbor-cli` can't work on this `ValidationUserName` function. @Vad1mo @bupd

Roaster05:
Hey @JianMinTang @Vad1mo @bupd, 

I would like to work on this issue. I am thinking of adding the following symbols to the username validation pattern:

- **Period (`.`)**: Allows separation of words for better readability and supports email-like usernames.
- **Underscore (`_`)**: Widely used to enhance clarity and readability without introducing ambiguity.
- **Hyphen (`-`)**: Links multiple words, improving visual structure while maintaining compatibility with URLs and CLI.

Additionally, I propose enforcing a rule that disallows consecutive symbols in usernames. Valid usernames will start and end with alphanumeric characters and may have symbols in between, ensuring that patterns like `abc--bacd` are rejected, while formats like `xyz.symbol.abc` are accepted.

Vad1mo:
> Hey @JianMinTang @Vad1mo @bupd, 
> 
> 
> 
> I would like to work on this issue. I am thinking of adding the following symbols to the username validation pattern:
> 
> 
> 
> - **Period (`.`)**: Allows separation of words for better readability and supports email-like usernames.
> 
> - **Underscore (`_`)**: Widely used to enhance clarity and readability without introducing ambiguity.
> 
> - **Hyphen (`-`)**: Links multiple words, improving visual structure while maintaining compatibility with URLs and CLI.
> 
> 
> 
> Additionally, I propose enforcing a rule that disallows consecutive symbols in usernames. Valid usernames will start and end with alphanumeric characters and may have symbols in between, ensuring that patterns like `abc--bacd` are rejected, while formats like `xyz.symbol.abc` are accepted.
> 
> 

The rules on the CLI should simply match with the rules on Harbor Server side.

Roaster05:
[...]

Issue #195, opened 2024-09-18
Chore: Validate and Trim Spaces for All User Input Fields


bupd:
## Expected Outcome
- All input fields should properly validate and trim spaces.
- If an input field contains only spaces or empty it should trigger an appropriate error message.

You can make use of Validation method mentioned [here.](https://github.com/goharbor/harbor-cli/pull/66#pullrequestreview-2312492192)

JianMinTang:
Hi, @bupd, I'm interested in this issue and have two ideas to solve the input problem. 
I'm not sure which one is better. 
The first one is:
![harborcli-#195-1](https://github.com/user-attachments/assets/7c7a856e-4ff1-4099-8f56-17ddbdcb5d43)
The second one is: 
![harborcli-#195-2](https://github.com/user-attachments/assets/537fac7a-6a5e-48e0-a6dc-5ce511669592)
And I'm not clear about the rules for trimming the string of input, do I need to trim only the leading of spaces, only the trailing of spaces, both the leading and trailing of spaces? Or are there other rules?
I'd really appreciate any additional helpful ideas you have to share!

bupd:
Hello @JianMinTang, thanks for your interest in addressing this issue.
> And I'm not clear about the rules for trimming the string of input, do I need to trim only the leading of spaces, only the trailing of spaces, both the leading and trailing of spaces? Or are there other rules?

Regarding your question on trimming the input string: 
The Go method [`strings.TrimSpace`](https://pkg.go.dev/strings#TrimSpace) will indeed trim both leading and trailing whitespaces. This should work well for cleaning up the input.

Here's my suggested approach to handle this across all user input fields:

- Step 1: Use `strings.TrimSpace` to remove both leading and trailing spaces from the input.
- Step 2: Check if the resulting string is empty, and if so, return an appropriate error.
- Additional Step: You can also apply more formatting rules & appropriate error messages as needed for specific fields to ensure the input adheres to the expected API resource model.

For reference, please take a look at the implementation in [this PR review](https://github.com/goharbor/harbor-cli/pull/66#pullrequestreview-2312492192). You can replicate the pattern used there for all relevant input fields. That's exactly the kind of input validation we need to implement throughout.

Feel free to share any further thoughts or ideas you might have!

JianMinTang:
Hello @bupd, i appreciate your encouraging and helpful reply in this issue, i try to implement the pattern code.
```
huh.NewInput().
            Title("Server").
[...]

PR #239, opened 2024-11-02
fixed username validator

Roaster05:
The username validator previously implemented only allowed alphanumeric characters, which resulted in rejecting valid usernames accepted by the server, such as `harbor-cli`. This inconsistency also caused failures in login tests.

To resolve this, the current implementation of the validator has been aligned with the server's username validation criteria. You can view the server-side validator [here](https://github.com/goharbor/harbor/blob/9e55afbb9a124dbe47562d56769ad4e9f14cb5ed/src/server/v2.0/handler/user.go#L514-L518).

This PR references issue #234

### DIFF
--- a/pkg/utils/helper.go
+++ b/pkg/utils/helper.go
@@ -44,10 +44,9 @@ func FormatSize(size int64) string {
 	return fmt.Sprintf("%.2fMiB", mbSize)
 }
 
+// ValidateUserName checks if the username is valid by length and allowed characters.
 func ValidateUserName(username string) bool {
-	pattern := `^[a-zA-Z0-9]{1,255}$`
-	re := regexp.MustCompile(pattern)
-	return re.MatchString(username)
+	return len(username) >= 1 && len(username) <= 255 && !strings.ContainsAny(username, `,"~#%$`)
 }
 
 func ValidateEmail(email string) bool {
--- a/pkg/utils/helper.go
+++ b/pkg/utils/helper.go
@@ -46,6 +46,7 @@ func FormatSize(size int64) string {
 
 // ValidateUserName checks if the username is valid by length and allowed characters.
 func ValidateUserName(username string) bool {
+	username = strings.TrimSpace(username)
 	return len(username) >= 1 && len(username) <= 255 && !strings.ContainsAny(username, `,"~#%$`)
 }
 


pkg/utils/helper.go
```diff
@@ -44,10 +44,9 @@ func FormatSize(size int64) string {
 	return fmt.Sprintf("%.2fMiB", mbSize)
 }
 
+// ValidateUserName checks if the username is valid by length and allowed characters.
 func ValidateUserName(username string) bool {
-	pattern := `^[a-zA-Z0-9]{1,255}$`
-	re := regexp.MustCompile(pattern)
-	return re.MatchString(username)
+	return len(username) >= 1 && len(username) <= 255 && !strings.ContainsAny(username, `,"~#%$`)
```
JianMinTang, line 49: Yes, it meets the username validators, but it also satisfy certain requirements outlined in the issue #195.

Roaster05: Just to confirm, should I add space trimming and an empty-check to the username validation here to fully meet issue #195 requirements, or handle that separately?

JianMinTang: I think it would be better to validate and trim spaces in the input fields here.

Roaster05: @JianMinTang @bupd  is it fine now?

bupd (approved):
LGTM

Vad1mo:
reverting back as test fail.

pkg/utils/helper.go
```diff
@@ -44,10 +44,10 @@ func FormatSize(size int64) string {
 	return fmt.Sprintf("%.2fMiB", mbSize)
 }
 
+// ValidateUserName checks if the username is valid by length and allowed characters.
 func ValidateUserName(username string) bool {
-	pattern := `^[a-zA-Z0-9]{1,255}$`
-	re := regexp.MustCompile(pattern)
-	return re.MatchString(username)
+	username = strings.TrimSpace(username)
```
JianMinTang, line 49: I believe this code may not work as expected. If you enter `harbor-cli` with extra spaces, the `strings.TrimSpace(username)` doesn't properly store the value in the `login.LoginView` variable. Could you please fix this? @Roaster05

bupd (commented):
.

PR #241, opened 2024-11-03
Support yaml output

### LINKED ISSUE #240: Harbor CLI does not support yaml output
This issue presists for all commands. yaml should be supported.
```bash
❯ ./harbor-dev registry list -h
list registry

Usage:
  harbor registry list [flags]

Flags:
  -h, --help            help for list
      --page int        Page number (default 1)
      --page-size int   Size of per page (default 10)
  -q, --query string    Query string to query resources
      --sort string     Sort the resource list in ascending or descending order

Global Flags:
      --config string          config file (default is $HOME/.harbor/config.yaml) (default "/home/bupd/.harbor/config.yaml")
  -o, --output-format string   Output format. One of: json|yaml
  -v, --verbose                verbose output

```

## This should be yaml but we get json
```bash
❯ ./harbor-dev registry list -o yaml
{
  "Link": "",
  "XTotalCount": 8,
  "Payload": [
    {
      "creation_time": "2024-10-07T13:27:37.749Z",
      "credential": {},
      "id": 2,
      "name": "8gears.container-registry.com",
      "status": "healthy",
      "type": "harbor",
      "update_time": "2024-10-07T13:27:45.201Z",
      "url": "https://8gears.container-registry.com"
```

We can take some functionalities from this PR #74.

Standing-Man:
Fixed #240

Support yaml output for the command `registry list`


bupd:
Every command need to have support for yaml output.

cmd/harbor/root/registry/list.go
```diff
@@ -23,10 +23,14 @@ func ListRegistryCommand() *cobra.Command {
...
 			FormatFlag := viper.GetString("output-format")
-			if FormatFlag != "" {
+			if FormatFlag == "json" {
 				utils.PrintPayloadInJSONFormat(registry)
 				return
 			}
+			if FormatFlag == "yaml" {
+				utils.PrintPayloadInYAMLFormat(registry)
+				return
+			}
```
bupd, line 33: Produce an error message to user when using flags other than json or yaml. 

bugInlisting

bupd (changes requested):

cmd/harbor/root/registry/list.go
JianMinTang: I have finalized the subcommand `list` for `project`, `registry`, and `user`. I would like to confirm that this is accurate before moving on to support the remaining commands.

JianMinTang: Should I use a generic function to unify the output in both YAML and JSON formats?

bupd (changes requested):
@JianMinTang thanks for the contribution. I found some commands still need additional formats. adding formats for those commands would be better.
 
## commands need to be covered are 
- artifact 
     - info
     - tags (these have issue with output)


```bash
❯ ./harbor-dev artifact tags list
INFO[0003] [0xc000695e60]

❯ ./harbor-dev artifact tags list -o yaml
INFO[0002] [0xc000034ba0]
```

- project 
    - view (also lacks table format)

- registry
    - info (lacks table format)
  
- repo
    - info (also lacks table format)

JianMinTang:
Hi @bupd, I am a bit confused about the difference between list and info under the same command. If there is any difference in the information they display, could you clearly explain the distinction between them?

bupd:
> If the user wants to view detailed information about a specific repository, they should use the command harbor repo info to see information about that particular repository.

`harbor repo view <project_name>/<repo_name>`

we can rename the `repo info` to `repo view` this might be clear than info.

bupd:
SO, @JianMinTang 

`repo list` - will list repos under a single project
`catalog` - will list all repos 
`repo view` - will show details about a single repo

JianMinTang:
Hi @bupd, I’ve added the table format and YAML/JSON output to some commands, following your suggestions. Thank you so much for your valuable input!

bupd (commented):
.

PR #248, opened 2024-11-12
Autogenerate credential name

Althaf66:
This will autogenerate the credential name in the format username@server.


bupd:
closing in favor of #250

Issue #249, opened 2024-11-12
Make tests not Dependent on Demo Server


bupd:
Instead of running test against the "demo.goharbor.io".  Create a dagger service running harbor and run tests against it.

This way no external service is required & not need to be maintained. 

This should be the way.

bupd:
Yes

PR #252, opened 2024-11-14
Add '--force' flag to delete non-empty projects

### LINKED ISSUE #243: Force deletes non empty projects
Projects that have artifacts can be deleted before the artifact is deletes. 

It would be good to have an `--force` option on delete that will first delete the artifacts and then the project. 

image

adwait-godbole:
Fixes #243 

[Screencast from 2024-11-14 17-03-59.webm](https://github.com/user-attachments/assets/587a9161-f8a9-4de2-a82b-a9b976662e5c)

Hi @Vad1mo, I tried fixing the problem and it is working fine on my machine as you can see in the above video. Let me know if any improvements should be done to do this approach as it is my very first time contributing to Harbor. Thanks for this amazing project!


Vad1mo:
@adwait-godbole can you rebase your changes on latest main? I think then the tests should work

adwait-godbole:
@bupd can you please review this one as well.  I think  the issue you have pointed out at #249 might be the reason that the tests are failing. Am I correct? Or is it something else?

bupd (approved):
LGTM!

PR #266, opened 2024-11-26
Better errormessage

Althaf66:
changed error messages for registry

### DIFF
--- a/pkg/errors/registry_error.go
+++ b/pkg/errors/registry_error.go
@@ -1,0 +1,50 @@
+package errors
+
+import (
+	"github.com/goharbor/harbor-cli/pkg/utils"
+)
+
+func getErrorMessage(statusCode int64, action string) string {
+	switch statusCode {
+	case 400:
+		if action == "create" {
+			return "Give adequate information to create a registry."
+		}
+	case 401:
+		return "Registry Unauthorized"
+	case 403:
+		return "Registry Forbidden"
+	case 404:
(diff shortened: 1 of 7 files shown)


pkg/errors/registry_error.go
bupd, line 1: simplify this into a simple function

pkg/utils/error.go
```diff
@@ -1 +1,18 @@
...
+	re := regexp.MustCompile(`\[(\d{3})\]`)
+	matches := re.FindStringSubmatch(err.Error())
+	var statusCode int64
+
+	if len(matches) > 1 {
+		stringCode := matches[1]
+		statusCode, _ = strconv.ParseInt(stringCode, 10, 64)
+	}
+	return statusCode
+}
```
bupd, line 18: implement a better way to get the status code. since regex can fail in different scenarios.

bupd (changes requested):

pkg/errors/registry_error.go
bupd: also remove duplicate functions.

pkg/utils/error.go
Althaf66: I think we only have the option to extract the status code from the error message.

Vad1mo (approved):
LGTM

bupd:
Closing this. Since, We need a better way to handle error messages.

errors should be handled on the place it created and should not be separated from the application logic.

PR #268, opened 2024-11-27
feature: Support YAML output for schedules

Standing-Man:
Add YAML format output in the `schedule` command

fixed #241


bupd (approved):
LGTM

PR #271, opened 2024-11-28
Fix Validation Domain for login command

### LINKED ISSUE #269: Add Validation for Server in login
Example

below should not work
check whether server is actually a server

Standing-Man:
fixed #269.

Add the Validation function


Vad1mo:
What happens for those use cases:

- subdomain.domain.com:port
- 127.0.0.1 and 127.0.0.1:8080
-

Vad1mo (changes requested):
What happens for those use cases:

- subdomain with prot -> subdomain.domain.com:**port**
- ip -> 127.0.0.1 and 127.0.0.1:8080
- top level domain -> registry.com

Standing-Man:
Hi @Vad1mo, I gave another validation method to support subdomains with or without ports, IP addresses, and top-level domains. Additionally, validated that the top-level domain of the domain name has a minimum length of at least two characters.

bupd (changes requested):
```bash
❯ ./harbor-dev login
┃ Server
┃ Server address eg. demo.goharbor.io
┃ > https://127.0.0.4:50

  User Name
  > lsadkjf

  Password
  >

  Name of Credential
  Name of credential to be stored in the harbor config file.
  > lsadkjf@1234-50

 * invalid domain format: too many colons
```

the above is a valid domain

## TIP
@Standing-Man, Instead of checking if the domain is valid. check only if its invalid by adding only these rules
Incorrect Domain Names:
```
http:// (missing domain)
https:// (missing domain)
http://example (missing TLD, not valid)
http://127.0.0.256 (invalid IP, exceeds valid octet range)
http://my_site.local (invalid local domain, underscores are not allowed in domain names)
https://example..com (double period, invalid syntax)
https://.example.com (leading dot in the domain, invalid)
http://-example.com/ (hyphen at the start, invalid)
https://example.c (single character TLD, invalid)
http://192.168.0.999 (invalid IP, octet exceeds range)
```
- Domain Names with Special Characters (Not Valid)
- http://example (missing TLD, not valid)

Standing-Man:
Hi @bupd, i'm confused about issue #297 and how to reproduce this issue. 
```
harbor login registry.bupd.xyz -> login successfully and store in config
harbor login http://registry.bupd.xyz/ -> failed to log in
```
or just
```
harbor login http://registry.bupd.xyz/  -> failed to pass the validation of server address
```
Anyway, i added more restrictions to validate server address and you can try to check if issue #297 solved.

bupd:
@Standing-Man server address with leading slash result in failed login. We should fix it.

pkg/utils/helper.go
```diff
@@ -39,6 +41,124 @@ func FormatUrl(url string) string {
 	return url
 }
 
+// ValidateDomain validates subdomain, IP, or top-level domain formats
+func ValidateDomain(domain string) error {
+	url := FormatUrl(domain)
```
bupd, line 46: In FormatUrl function. I would suggest adding the removal of leading slash.

bupd (changes requested):

pkg/utils/helper.go
Standing-Man: image

The URL `http://registry.bupd.xyz/` is still working for me, so I'm a bit unsure about what might be going wrong.

bupd (changes requested):
Thanks for your contributions

Please do Rebase.

PR #274, opened 2024-11-29
Add jq to container image

Jellyfrog:
It makes it easier to parse out fields, for example to get a registry id when doing something like this inside a container:
```shell
./harbor registry create --name docker.io ...
REGISTRY_ID=$(./harbor registry list -q "name=docker.io" -o json | jq '.Payload[].id')
./harbor project create docker.io --proxy-cache=true --registry-id "$REGISTRY_ID"
```

### DIFF
--- a/dagger/main.go
+++ b/dagger/main.go
@@ -147,6 +147,7 @@ func (m *HarborCli) PublishImage(
 
 		ctr := dag.Container(dagger.ContainerOpts{Platform: dagger.Platform(os + "/" + arch)}).
 			From("alpine:latest").
+			WithExec([]string{"apk", "add", "--no-cache", "jq"}).
 			WithFile("/harbor", builder.File("./harbor")).
 			WithEntrypoint([]string{"./harbor"})
 		releaseImages = append(releaseImages, ctr)


bupd:
Thanks for your contribution @Jellyfrog .

Not sure if we need the jq inside the container itself.

cc @Vad1mo

Jellyfrog:
The problem is that you can't rely on only `harbor`-commands to register a project as a proxy cache since you can't get only the `id` field of a registry with any `harbor` command. So this eases the pain until something like `./harbor registry list -q "name=docker.io" --field id` is available`

Vad1mo:
on one hand yes on the other no. 
I think we woud need to have two images one scratch that you run on your cli and the other one e.g. debug that you can run in ci/cd that has a shell, where you can do some operations inside. 

Mybe we  will leave it just with one debug version

Jellyfrog:
@Vad1mo maybe thats a good idea! if you release a 0.0.2 image then it's possible to do this now which is not that terrible: 
`./harbor registry view docker.io -o yaml | grep " id: " | awk '{print $2}'`

PR #275, opened 2024-11-30
Skip Loginview create when credentials already exist

### LINKED ISSUE #267: Storing Plaintext in Config Files is Insecure
image
We might consider using secure method to store passwords in the config file, and it would be helpful to include instructions on how to use `--config` with the `login` command in the README and CLI manual.

Next, could we enable logging in through the config file instead of manually entering credentials each time?

qcserestipy:
Fixes #267 

# Check for Existing Credentials and Bypass Login View Creation

1. Implemented a verification step to detect existing Harbor credentials in the current configuration.
2. If valid credentials are found, the creation of a new login view is skipped.
3. Enhances login efficiency by reusing stored credentials and reducing unnecessary prompts.

ToDo: Add base64 encoding for password in config file?


qcserestipy:
To resolve the failing tests I will introduce a mock Keyring since the container has no keyring.

Vad1mo (approved):

bupd (commented):
There should be an easy way to change the login credentials directly. 

- Changing login credentials should not need users to change the config.yaml

qcserestipy:
@bupd Do you thing it would come in handy to create a config subcommand that lets us manipulate the config file?
Such that a user could do something like `harbor config set Credential.Password <password>` to update a field in the config file.

qcserestipy:
@bupd  I added the config sub command and several tests for its functionality. I added some screenshots:

### Listing Config Items and Setting them
Screenshot 2024-12-22 at 13 53 59

### Clearing config items
Screenshot 2024-12-22 at 13 59 58

### Retrieving config items
Screenshot 2024-12-22 at 13 56 46

### Setting credentials and credentialname supported with --name flag
Screenshot 2024-12-22 at 13 56 01

Screenshot 2024-12-22 at 13 56 11

Screenshot 2024-12-22 at 13 54 25

qcserestipy:
@bupd There is still a sign off error in one of the commits. Unfortunately, I would have to rebase over 34 commits from main. Is there an easier way to do this?

bupd:
Try squashing the commits

bupd:
@qcserestipy did you also add changing between different login accounts via changing the credential name command. 

Which utilizes the current credential name.

qcserestipy:
@bupd I went through the rebase and it worked with a little bit of manual work.
Concerning your second question can you give me an example what you mean?

You can set the `current-credential-name` with the command too:

Screenshot 2024-12-22 at 14 33 19

Standing-Man:
Hi @qcserestipy and @bupd, If the command format is as follows, it might be even better.
- `harbor context list` -> list all context
- `harbor context create` -> create a new context for now 
- `harbor context view` -> view current context
-  `harbor context delete` -> delete context 
- ....

This is just a suggestion I have for this command. I'm not sure if it would be more user-friendly, but I’d love to hear your thoughts!

bupd (changes requested):
There is some weird behaviour for me while testing this.
1. config file gets resurrected automagically with previous credentials.

Attaching video for reference.

https://github.com/user-attachments/assets/73ffbf35-ccdd-460d-97b1-88cf581ef8a9


2. I can't create a new login.
3. Main focus of config command should be - allowing users to change the current credential or login with new ones. Since people should have ability to login with different harbor instances, directly from the CLI

bupd:
@qcserestipy to reduce the complexity. 
I would suggest using the login command to switch between contexts and also offer. to use new login too.

qcserestipy:
@bupd I made the changes that you suggested. Users can now change the context via the `--name` flag of the login command.

doc/man-docs/man1/harbor-config-set.1
```diff
@@ -0,0 +1,57 @@
...
+
+
+.SH EXAMPLE
+.EX
+
+  # Set the current credential's password
+  harbor config set credentials.password myNewSecret
+
+  # Set a credential's password by specifying the credential name
+  harbor config set credentials.password myNewSecret --name harbor-cli@http://demo.goharbor.io
```
Vad1mo, line 47: the example should be a bit different here. eg. 

```
credentials:
    - name: admin@demo-goharbor-io

```

Vad1mo (commented):
deleting the whole context entry woudl be nice too.

Vad1mo:
We did some testing and in generally it looks good to me.

cmd/harbor/root/config/delete.go
```diff
@@ -0,0 +1,244 @@
...
+		Short: "Delete (clear) a specific config item",
+		Example: `
+  # Clear the current credential's password
+  harbor config delete credentials.password
+
+  # Clear a specific credential's password using --name
+  harbor config delete credentials.password --name admin@http://demo.goharbor.io
+
+  # Clear all credentials and the current credential
+  harbor config delete --all
```
bupd, line 44: can we make  the flag `--current` instead of `--all` since all would mean delete all of the credentials.
```suggestion
  # Clear the current credential
  harbor config delete --all
```

cmd/harbor/root/config/update.go
```diff
@@ -0,0 +1,267 @@
...
+	"github.com/spf13/cobra"
+)
+
+// SetConfigItemCommand creates the 'harbor config set' subcommand,
+// allowing you to do: harbor config set <item> <value>.
+func SetConfigItemCommand() *cobra.Command {
+	var credentialName string
+
+	cmd := &cobra.Command{
+		Use:   "set <item> <value>",
```
bupd, line 33: Instead of set we should call it update here.

```suggestion
		Use:   "update <item> <value>",
```

cmd/harbor/root/config/delete.go
bupd: BTW can we just set to delete the current config with both username and password by default.

bupd (changes requested):
Thanks for the contributions 

Left some minor nitpicks.

bupd:
Overall Looks Good to me. But the only thing stopping us from merging is the git commit history of this PR. which includes duplicate commits which is already in main branch.

If you are done with the fixes let me know. I am ready to help. We can do a cherry-pick and create squashed new PR only with necessary commits. 

Hope this helps. Thanks

cmd/harbor/root/config/delete.go
qcserestipy: What do you mean with `by default`? It is implemented such that the `--current` flag deletes the whole credentials list item. Is that what you mean or can you give me an example?

qcserestipy:
@bupd Thank you for the extensive reviews! I am happy to hear this code is nearly ready for being merged. From my side I would be ready for making a squashed PR with necessary commits.

qcserestipy:
Since #372 was merged we can close this one.

Issue #94, opened 2024-05-30
📌 Tracker: Missing and To-Be-Added CLI Commands


bupd:
## Objective: 
Track the missing commands in the CLI and ensure their implementation.
Current focus is to add more commands which includes.

1. immutability tag rules
2. set quotas
3. set replication policies
4. create retention policies
5. vulnerability scanning policy 
6. webhooks

and more commands mentioned below.

### 🚧 Missing Commands

| Command         | Description                       | Priority | Status |
| --------------- | --------------------------------- | -------- | ------ |
| **Member**       | Manage project Members             | High     | #85   |
| **Quota**       | Manage project quotas             | High     | #97   |
| **Robot**       | Manage robot accounts in projects | High     | #100  |
| **Replication** | Manage replication policies       | High     | done  |
| **Retention**   | Manage retention policies         | High     | done  |
| **Ping**        | Ping the API server               | Low      | done  |
| **Webhook**       | Manage project webhooks     | Medium | done  |
| **Security**      | Manage system security settings  | Medium | done  |
| **Configuration** | Manage system configurations | Medium | done  |
| **System Config** | Manage internal system configurations | Medium | done | 
| **Audit**       | Manage and audit recent logs      | Medium   | to-do  |
| **Usergroup**   | Manage user groups                | Medium   | to-do  |

---

### 🛠️ Commands to Handle After Initial Implementation

| Command           | Description                   |
| ----------------- | ------------------------------------- |
| **LDAP**          | Integrate with LDAP   |
| **Preheat**       | Manage preheat policies and instances |
| **Permissions**   | Manage system and project permissions |

---

### 📝 Implementation Steps

   - [x] Implement Member command.
   - [x] Implement Quota command.
   - [x] Implement Robot command.
   - [x] Implement Audit command.
   - [x] Ping
   - [x] Implement Replication command.
   - [x] Webhook
   - [x] Security
   - [x] Configuration (ongoing)
   - [x] System Config (ongoing)
   - [x] LDAP
   - [ ] Implement Retention command. ([ongoing #364](https://github.com/goharbor/harbor-cli/pull/364))
   - [ ] Implement Usergroup command.
   - [ ] Preheat
   - [ ] Permissions
   - [ ] Add Unit, End to end tests for the commands.

---

### 📌 Issue Tracking and Progress

- Create individual GitHub issues for each command implementation.
- Reference this master issue in each individual issue to maintain a coherent structure and clear dependencies.

---

[...]

PR #285, opened 2024-12-17
chore: move the ping to health handler

Standing-Man:
fix #211 
part of #94 
~~move the ping command handler and verify the API server first, before checking the status of the Harbor components.~~

Edit:
Add a Ping step to the health command that pings the server before checking its health status. Removing the need for separate ping command.


bupd (approved):
lgtm

PR #289, opened 2024-12-20
User subcommand

Standing-Man:
Add additional subcommand for user command
- `harbor user reset [username]` -> for reset the user's password
- `harbor user update [username]` -> for update the user's profile


Standing-Man:
Hi @Vad1mo, I’ve added two additional subcommands under `user` and wanted to check if it would be helpful to include a subcommand for setting a user’s role to Administrator or revoking the Administrator role.

bupd:
@Standing-Man Thanks for your contributions.

cmd/harbor/root/user/update.go
```diff
@@ -0,0 +1,72 @@
...
+			err = api.UpdateUserProfile(updateView, UserId)
+			if err != nil {
+				logrus.Errorf("failed to update user's profile: %v", err)
+			}
+		},
+	}
+	flags := cmd.Flags()
+	flags.StringVarP(&opts.Comment, "comment", "m", "", "Comment of the user")
+	flags.StringVarP(&opts.Email, "email", "e", "", "Email of the user")
+	flags.StringVarP(&opts.Realname, "realname", "r", "", "Realname of the user")
```
bupd, line 69: Add a flag to use userID directly
```suggestion
	flags.StringVarP(&opts.Realname, "realname", "r", "", "Realname of the user")
	flags.StringVarP(&UserID, "userID", "id", "", "ID of the user")
```

pkg/api/user_handler.go
```diff
@@ -36,6 +37,26 @@ func CreateUser(opts create.CreateView) error {
...
+		Password: &models.PasswordReq{
+			OldPassword: resetView.OldPassword,
+			NewPassword: resetView.NewPassword,
+		},
+	})
+
+	if err != nil {
+		return err
+	}
+	log.Infof("User reset password successfully with id %d", userId)
```
bupd, line 56:
```suggestion
	log.Infof("User password reset  successfully with id %d", userId)
```

pkg/api/user_handler.go
```diff
@@ -106,3 +127,36 @@ func GetUsersIdByName(userName string) (int64, error) {
...
+
+	_, err = client.User.UpdateUserProfile(ctx, &user.UpdateUserProfileParams{
+		UserID:  userId,
+		Profile: profile,
+	})
+	if err != nil {
+		return err
+	}
+
+	log.Info("user's profile updated successfully")
```
bupd, line 159: Maintain log messages consistently

bupd (changes requested):

cmd/harbor/root/user/reset.go
```diff
@@ -0,0 +1,42 @@
...
+		Short: "reset user's password",
+		Long:  "reset user's password by username",
+		Args:  cobra.MaximumNArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			var err error
+			var userId int64
+			resetView := &reset.ResetView{}
+
+			if len(args) > 0 {
+				userId, _ = api.GetUsersIdByName(args[0])
```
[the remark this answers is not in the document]
Standing-Man: I modified the error handling of `api.GetUsersIdByName` because it doesn't raise an error even when the userId is not found.

Standing-Man:
enhance :
- add `--userID` flag on update subcommand  
- fix the error handler of `api.GetUsersIdByName`
- supplement the descriptions for reset and update
- maintain log messages consistently

cmd/harbor/root/user/reset.go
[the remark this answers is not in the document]
bupd: Eventhough if we have problems with api. if you can fix then fixit. Otherwise It is Still its better to handle the error, Instead of ignoring it.

bupd:
@Standing-Man for the suggestions that are added mark the conversations as resolved. 
Thanks

cmd/harbor/root/user/reset.go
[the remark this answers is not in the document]
Standing-Man: sure, i have solved this problem on new commit.

Standing-Man:
Thank you for your reminder @bupd. I have marked these conversations as resolved.

Standing-Man:
Hi @Vad1mo and @bupd, ready to review.

pkg/views/user/reset/view.go
```diff
@@ -0,0 +1,64 @@
...
+				}),
+			huh.NewInput().
+				Title("Comfirm Password").
+				EchoMode(huh.EchoModePassword).
+				Value(&resetView.Comfirm).
+				Validate(func(str string) error {
+					if resetView.Comfirm != resetView.NewPassword {
+						return errors.New("passwords do not match, please try again")
+					}
+					return nil
```
bupd, line 56: Add validation here for checking if old and new passwords match

make sure old vs newe match pass

The above should result in and error.
Please add that check

cmd/harbor/root/user/update.go
```diff
@@ -0,0 +1,75 @@
...
+			if err != nil {
+				logrus.Errorf("failed to update user's profile: %v", err)
+			}
+		},
+	}
+	flags := cmd.Flags()
+	flags.Int64VarP(&UserID, "userID", "d", 0, "ID of the user")
+	flags.StringVarP(&opts.Comment, "comment", "m", "", "Comment of the user")
+	flags.StringVarP(&opts.Email, "email", "e", "", "Email of the user")
+	flags.StringVarP(&opts.Realname, "realname", "r", "", "Realname of the user")
```
bupd, line 72:
```suggestion
	flags := cmd.Flags()
	flags.Int64VarP(&UserID, "userID", "", 0, "ID of the user")
	flags.StringVarP(&opts.Comment, "comment", "", "", "Comment of the user")
	flags.StringVarP(&opts.Email, "email", "", "", "Email of the user")
	flags.StringVarP(&opts.Realname, "realname", "", "", "Realname of the user")
```

Please remove shortforms, as this might mislead users. 
also removing this will maintain consistency across the CLI.

cmd/harbor/root/user/reset.go
```diff
@@ -0,0 +1,46 @@
...
+			}
+
+			reset.ResetUserView(resetView)
+
+			err = api.ResetPassword(UserID, *resetView)
+
+			if err != nil {
+				logrus.Errorf("failed to reset user's password: %v", err)
+			}
+
```
bupd, line 40: Also add flag here to enable reset the user's password via UserID directly.

bupd (changes requested):
Thanks for your contributions

I have left some suggestions for improvements.

cmd/harbor/root/user/update.go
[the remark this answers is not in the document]
bupd: since harbor UI allows this

pkg/views/user/reset/view.go
Standing-Man: Hi @bupd, i have alredy added the checking if  the new password is same as the old's, but i'm confused about what this snapshot mean.

Standing-Man:
Hi @bupd, When I add a new command, I need to update the documentation. However, running `dagger call run-doc` and `go run doc.go` does not generate the corresponding command file.

pkg/views/user/reset/view.go
bupd: Password and confirm password should be the same.

Standing-Man: image
I had added a check to verify if the confirmation password matches the new password.

bupd:
The git history of this is not right. 

Can you fix these ones. @Standing-Man

Standing-Man:
> The git history of this is not right.
> 
> Can you fix these ones. @Standing-Man
Hi @bupd, I'm confused where the git history is wrong. Could you please point out?

bupd (changes requested):
Please fix the git commit history. only update files that are necessary

image

image

Standing-Man:
Hi @bupd, I resolved the git history issue by submitting PR #330.

PR #290, opened 2024-12-22
Default Page size

### LINKED ISSUE #244: Ability to configure default page size
I would be beneficial to be able to configure the default page size for different harbor cli operations.

### LINKED ISSUE #261: Add config option for defaults in cli
We need to add a configuration option in the existing CLI config file to allow users to set a default page size. and to set defaults.

Roaster05:
Fix #244 
also fix #261 
## Description  
This PR enhances the handling of the `page-size` parameter by introducing a hierarchical priority mechanism and integrating it into the command execution flow.  

### Hierarchy for `page-size` Resolution  
1. **Flag**: The value provided via the command flag takes the highest priority.  
2. **ResolvePageSize Utility**: If no flag is provided, the utility determines the `page-size` in the following order:  
   - Value from the configuration file.  
   - Value from the `HARBOR_CLI_DEFAULT_PAGE_SIZE` environment variable.  
   - Hardcoded default value (`x`).  

### Key Changes  
1. **Command Execution**:  
   - Updated `list` commands to prioritize `page-size` values using the hierarchy mentioned above.  
2. **Config Updates**:  
   - Introduced a new parameter for `page-size` in the configuration.  
   - Added initialization for this parameter in the relevant configuration files.

### DIFF
--- a/cmd/harbor/root/labels/list.go
+++ b/cmd/harbor/root/labels/list.go
@@ -16,6 +16,9 @@ func ListLabelCommand() *cobra.Command {
 		Use:   "list",
 		Short: "list labels",
 		Run: func(cmd *cobra.Command, args []string) {
+			if !cmd.Flags().Changed("page-size") {
+				opts.PageSize = utils.ResolvePageSize(20)
+			}
 			label, err := api.ListLabel(opts)
 			if err != nil {
 				log.Fatalf("failed to get label list: %v", err)
--- a/cmd/harbor/root/project/list.go
+++ b/cmd/harbor/root/project/list.go
@@ -20,6 +20,9 @@ func ListProjectCommand() *cobra.Command {
 		Use:   "list",
 		Short: "list project",
 		Run: func(cmd *cobra.Command, args []string) {
+			if !cmd.Flags().Changed("page-size") {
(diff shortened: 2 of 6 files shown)


bupd (changes requested):
Thanks for your contributions. Here is one change.

Instead of giving different page sizes at different places. 

We should have single default page size. that is configurable.

Roaster05:
@bupd It seems there was some confusion regarding the value passed in the `ResolvePageSize(x int64) int64` function. 

To clarify:
- The parameter value `x` is only used when:
  1. There is no configuration available in the config file.
  2. No environment variable is explicitly set for the default page size.

In such cases, the function assigns a hardcoded value to serve as the default. Previously, this hardcoded value varied across different files (`10` or `20`), leading to inconsistency.

To address this:
- The current implementation uses the parameter `x` to provide the hardcoded default value when none of the config or environment variables are set.
- If we want to enforce a consistent hardcoded value across all places, we can eliminate the parameter `x` entirely and define the hardcoded value directly within the function. This ensures consistency and simplifies the function signature.

Let me know if you want to proceed with this approach or need further clarification!

bupd:
Thanks GPT

Issue #293, opened 2024-12-28
Unexpected Behaviour: Should not allow extra Args to project list


bupd:
## To Reproduce this error
Run `harbor project list <anything>`

## Expected Behaviour
Run `harbor project list <anything>`

The above command should produce **error** as shown below.
```bash
❯ ./harbor-dev project list create
Error: accepts 0 arg(s), received 1
```

## Current Behaviour
Run `harbor project list <anything>` works fine.
```bash
❯ ./harbor-dev project list create     #this should not work
┌──────────────────────────────────────────────────────────────────────────────────────────────┐
│  ID      Project Name          Access Level  Type          Repo Count    Creation Time       │
│ ──────────────────────────────────────────────────────────────────────────────────────────── │
│  590     11aaaa                private       proxy cache   0             1 day ago           │
│  516     aaa                   public        project       3             15 day ago          │
│  523     af1-test              private       project       1             15 day ago          │
│  496     aloha-test            private       project       2             18 day ago          │
```

### Note
Review all the commands that does not require args and set their arguments to 0. if they are not already specified.

shikharburman:
Hello, I would like to work on this issue.

rizul2108:
@bupd we can close this issue now

PR #295, opened 2024-12-29
fixed extra args in cmds

Roaster05:
This PR fixes issue #293 and extends its coverage to other commands by ensuring that an error message is sent if any extra arguments are provided to commands that do not require them.

Current behavior:

```
./harbor-cli label create newLabel
FATA[0000] Error: accepts 0 arg(s), received 1: [newLabel]

./harbor-cli project list extra args
FATA[0000] Error: accepts 0 arg(s), received 2: [extra args]
```

The fix adds a PreRun step before the command is executed to exit if any arguments are provided.

### DIFF
--- a/cmd/harbor/root/health.go
+++ b/cmd/harbor/root/health.go
@@ -3,6 +3,7 @@ package root
 import (
 	"github.com/goharbor/harbor-cli/pkg/api"
 	"github.com/goharbor/harbor-cli/pkg/views/health"
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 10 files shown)


bupd:
> The fix adds a PreRun step before the command is executed to exit if any arguments are provided.

We acutally don't need it. It adds unnecessary complexity

Roaster05:
> > The fix adds a PreRun step before the command is executed to exit if any arguments are provided.
> 
> We acutally don't need it. It adds unnecessary complexity

Yes, we can also just add `Args: cobra.NoArgs`, in the cmd definition, but this will throw an error like `Error: unknown command "newLabel" for "harbor label create"`. However, if we need to have our custom error message like the one mentioned in #293, I thought of adding a PreRun. We can also simply raise that error in the Run step itself based on the length of args

PR #303, opened 2025-01-03
docs: update README to include directory change after cloning

rizul2108:
there was this issue that there was no command to move to cloned directory after cloning so this PR fixes that 
Earlier it gave error like this 
```
rizul@rizu:~$ go build -o harbor-cli cmd/harbor/main.go
package cmd/harbor/main.go is not in std (/snap/go/10748/src/cmd/harbor/main.go)
```


rizul2108:
@bupd i have to create issue also for this PR ?

Issue #306, opened 2025-01-09
Project Create cmd not working


bupd:
## Description
`project create` command fails with the following error.

```
~/s/code/OSS/harbor-cli/main main*
❯ ./harbor-dev project create
credential with name '' not found%
```

Other commands seems to work fine such as list, view. the problem is only with the `harbor project create` for now.

rizul2108:
does this command take input of name from the user like for `label create` command ?

rizul2108:
i found the error but i dont know how many places we will have to fix that

rizul2108:
error is in this below function
https://github.com/goharbor/harbor-cli/blob/8dd8abb21c626fca3cb8488d027bb71c120967b0/pkg/utils/config.go#L171C1-L180C1

here we are creating a new viper client using viper.new but instead everywhere v as new client has been used viper should be used as it acts a global client for whole project and this same thing has been done in many places in this file so should I make PR fixing all those or should I fix only those lines concerned with the `project create` command

rizul2108:
@bupd can you please review the PR ?

PR #307, opened 2025-01-09
fix: Project create cmd fix

rizul2108:
This PR fixes issue #306 

There was error in this below function 
https://github.com/goharbor/harbor-cli/blob/8dd8abb21c626fca3cb8488d027bb71c120967b0/pkg/utils/config.go#L171C1-L180C1
here we were creating a new viper client using `viper.new` and everywhere v as new client has been used instead viper should be used as it acts a global client for whole project and if v was used it acted as a local new client for that function only and because of this new client the `current-credential-name` string in below line was found to be empty as it had no context of the global configurations 

https://github.com/goharbor/harbor-cli/blob/8dd8abb21c626fca3cb8488d027bb71c120967b0/pkg/views/project/create/view.go#L25

So I removed `viper.new` and `v` and used viper directly in that function and now the create command works properly as expected


bupd:
@rizul2108 please do rebase

rizul2108:
@bupd this is okay?

bupd (approved):
lgtm

rizul2108:
Can you merge this pr @bupd

bupd:
fixes #306

Vad1mo (approved):

PR #314, opened 2025-01-19
feat: support labels add/remove on artifact

Standing-Man:
add subcommand `lables` to manage labels of an artifact

`harbor artifact labels add` -> Add a label of an artifact
`harbor artifact labels remove` -> Remove a label of an artifact


cmd/harbor/root/artifact/labels/add.go
```diff
@@ -0,0 +1,45 @@
...
+					logrus.Errorf("Failed to get this lable: %s", args[1])
+					return
+				}
+			} else {
+				projectName = prompt.GetProjectNameFromUser()
+				repoName = prompt.GetRepoNameFromUser(projectName)
+				reference = prompt.GetReferenceFromUser(repoName, projectName)
+				labelID = prompt.GetLabelIdFromUser(opts)
+			}
+			err = api.AddLabel(projectName, repoName, reference, labelID)
```
Standing-Man, line 35: Hi @bupd and @Vad1mo, `golangci-lint` detected duplicate code in the add and remove subcommands, but since they serve different purposes, I'm wondering if there's a recommended way to handle this. Any suggestions?
image
image

PR #324, opened 2025-02-06
add projectID flag option to delete project command

### LINKED ISSUE #309: Force Project deletion does not work by id
I was trying to force delete a project by ID and it failed with the below error message, when I tried to delete it by ID it worked. 

Harbor CLI should be able to cope with both id and name.

```
harbor-cli/bin on  main [$!] 
❯ ./harbor-cli project delete 560 --force
ERRO[0000] failed to list repositories: [GET /projects/{project_name}/repositories][404] listRepositoriesNotFound  &{Errors:[0x1400011c160]} 
ERRO[0000] failed to delete project: [GET /projects/{project_name}/repositories][404] listRepositoriesNotFound  &{Errors:[0x1400011c160]} 
```
```

harbor-cli/bin on  main [$!] 
❯ ./harbor-cli project delete devsecops --force
INFO[0000] Repository devsecops/test-image deleted successfully 
INFO[0000] Project devsecops deleted successfully     
```

adityachopra29:
# Which problem is this PR solving?
- fixes #309 

# Description of the changes
- add `projectID` flag to `delete` command in `project` command.
- add `projectID` parameter to all corresponding functions
- update project and repository handler functions accordingly

# How was this change tested?
- Running the delete command on one of the projects and verify it getting deleted:
Screenshot 2025-02-07 at 4 54 55 AM

### DIFF
--- a/cmd/harbor/root/project/delete.go
+++ b/cmd/harbor/root/project/delete.go
@@ -23,19 +23,20 @@ import (
 // DeleteProjectCommand creates a new `harbor delete project` command
 func DeleteProjectCommand() *cobra.Command {
 	var forceDelete bool
+	var useProjectID bool
 
 	cmd := &cobra.Command{
 		Use:   "delete",
-		Short: "delete project by name or id",
+		Short: "Delete project by name or ID",
 		Args:  cobra.MaximumNArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
 			var err error
 
 			if len(args) > 0 {
-				err = api.DeleteProject(args[0], forceDelete)
+				err = api.DeleteProject(args[0], forceDelete, useProjectID)
 			} else {
 				projectName := prompt.GetProjectNameFromUser()
-				err = api.DeleteProject(projectName, forceDelete)
+				err = api.DeleteProject(projectName, forceDelete, false)
 			}
 			if err != nil {
 				log.Errorf("failed to delete project: %v", err)
@@ -45,6 +46,7 @@ func DeleteProjectCommand() *cobra.Command {
 
 	flags := cmd.Flags()
 	flags.BoolVar(&forceDelete, "force", false, "Deletes all repositories and artifacts within the project")
+	flags.BoolVar(&useProjectID, "projectID", false, "If set, treats the provided argument as a project ID instead of a project name")
 
 	return cmd
 }
--- a/cmd/harbor/root/project/view.go
+++ b/cmd/harbor/root/project/view.go
@@ -32,16 +32,16 @@ func ViewCommand() *cobra.Command {
 		Args:  cobra.MaximumNArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
 			var err error
-			var projectName string
+			var projectNameOrID string
 			var project *project.GetProjectOK
 
 			if len(args) > 0 {
-				projectName = args[0]
+				projectNameOrID = args[0]
 			} else {
-				projectName = prompt.GetProjectNameFromUser()
+				projectNameOrID = prompt.GetProjectNameFromUser()
 			}
 
-			project, err = api.GetProject(projectName)
+			project, err = api.GetProject(projectNameOrID, false)
 			if err != nil {
 				log.Errorf("failed to get project: %v", err)
 				return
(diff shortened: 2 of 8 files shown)


adityachopra29:
@bupd I have made the changes. Plz have a look if anything needs to be changed.
A few points I wanted to discuss:
- In the client sdk for the operations on `repositories`, I found that all the querying methods only used projectName, unlike as provided in the api. Hence, I had to first use the updated getProject() method to first get the projectName from the projectID and then use it in the sdk methods. Is that the most optimal way or is there a better solution?
- Methods defined in /prompt are not changed yet, as that would require making change in many more functions. If this is the correct way of solving this issue, I will make those changes as well.
- A similiar issue as #309 is present for the rest of the repository operation commands (ie `view`, `list`) and project commands(such as `project view` ), as they also work only with `projectName`. Those changes can be made in this pr itself or later on, whatever is more suitable.

cmd/harbor/root/project/delete.go
```diff
@@ -45,6 +46,7 @@ func DeleteProjectCommand() *cobra.Command {
 
 	flags := cmd.Flags()
 	flags.BoolVar(&forceDelete, "force", false, "Deletes all repositories and artifacts within the project")
+	flags.BoolVar(&useProjectID, "projectID", false, "If set, treats the provided argument as a project ID instead of a project name")
```
bupd, line 49: I would suggest project ID to be a flag to which you can pass project ids. rather than having it as a boolean. since this would allow the args to always be projectname maintaining consistency. 

Thanks

cmd/harbor/root/project/view.go
```diff
@@ -32,16 +32,16 @@ func ViewCommand() *cobra.Command {
 		Args:  cobra.MaximumNArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
 			var err error
-			var projectName string
+			var projectNameOrID string
```
bupd, line 35: have it just projectName and have different var for projectID.

pkg/utils/helper.go
```diff
@@ -153,3 +153,8 @@ func PrintFormat[T any](resp T, format string) error {
 	}
 	return fmt.Errorf("unable to output in the specified '%s' format", format)
 }
+
+func CreateBoolPointer(b bool) *bool {
+    return &b
+}
+
```
bupd, line 160: remove this

pkg/api/repository_handler.go
```diff
@@ -51,11 +60,20 @@ func RepoView(projectName, repoName string) (*repository.GetRepositoryOK, error)
 	return response, nil
 }
 
-func ListRepository(projectName string) (repository.ListRepositoriesOK, error) {
+func ListRepository(projectNameOrID string, useProjectID bool) (repository.ListRepositoriesOK, error) {
 	ctx, client, err := utils.ContextWithClient()
 	if err != nil {
 		return repository.ListRepositoriesOK{}, err
 	}
+	projectName := projectNameOrID
```
bupd, line 68: wrong wording.

bupd (changes requested):
Thanks for you contribution

I have left a review on changes that can be made on this PR.

cmd/harbor/root/project/delete.go
adityachopra29: > I would suggest project ID to be a flag to which you can pass project ids. rather than having it as a boolean. since this would allow the args to always be projectname maintaining consistency.

@bupd 
So are you suggesting to make projectName an optional input for the delete command? Because I was under the understanding that we are giving projectID when we did not know the projectName.

Right now, when we don't give projectName, we enter the prompt mode and again choose projectName. So do we want to change that behavior when the projectID flag is given?

adityachopra29: @bupd Could you please clarify the above point? 

Also,
> args to always be projectname maintaining consistency.

Just for more clarification, I planned on adding/replacing the projectNameOrId flag to the rest of the commands as well. This way we will also have consistency.

bupd (changes requested):
merge conflicts!!

cmd/harbor/root/project/view.go
rizul2108: We should keep it as a single variable for consistency with Harbor API. Similar to how we retrieve a project by passing a string that could be either a name or an ID, using a single variable (projectNameOrID) maintains uniformity.

pkg/api/repository_handler.go
rizul2108: Can you please elaborate more what needs to be fixed here?

PR #360, opened 2025-03-22
feat: set `project list` cmd to list all projects by default

rizul2108:
# feat: update `project list` cmd to return all projects by default

## Summary
This PR updates the `project list` command to return **all projects** by default, even if no `--page-size` is specified. Previously, the command would only return **10 projects** by default due to the Harbor API's default behavior. This update ensures that the CLI will keep requesting additional pages until all projects are retrieved, unless a `--page-size` is explicitly provided.

## Changes
- Updated the `project list` command to:
  - If `--page-size` is **not provided**, the CLI will:
    - Set the `page-size` to `100` (maximum allowed by Harbor API).  
    - Keep requesting subsequent pages until all projects are retrieved.  
  - If `--page-size` is **provided**, it will use the specified value.  
- Optimized the pagination logic to minimize redundancy and improve efficiency.  
- Ensured consistent output formatting using `utils.PrintFormat`.

## Why
- The Harbor API defaults to returning **only 10 projects** per request when no `page-size` is specified.  
- This leads to incomplete results in the case of CLI unless the user explicitly sets a `--page-size` and `--page`.  
- This fix improves the user experience by returning **all projects** without manually adjusting the page size and page.  

## How
- Implemented a loop to handle paginated results:
  - Request projects with a `page-size` of `100`.  
  - Keep fetching until fewer than `100` projects are returned, indicating the last page.  
- Ensured that pagination works correctly even if `--page-size` is provided.  

## Checklist
- [x] Code tested locally  
- [ ] Updated documentation

### DIFF
--- a/cmd/harbor/root/project/list.go
+++ b/cmd/harbor/root/project/list.go
@@ -15,6 +15,7 @@ package project
 
 import (
 	"github.com/goharbor/go-client/pkg/sdk/v2.0/client/project"
+	"github.com/goharbor/go-client/pkg/sdk/v2.0/models"
 	"github.com/goharbor/harbor-cli/pkg/api"
 	"github.com/goharbor/harbor-cli/pkg/utils"
 	list "github.com/goharbor/harbor-cli/pkg/views/project/list"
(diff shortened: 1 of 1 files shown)


rizul2108:
@bupd should I update like this for all `list` commands?

bupd (changes requested):
please add screenshots.
Thanks

rizul2108:
image
image

The above one in both the images is for the older version when the default was 10, and the below one is an  updated `list` command that lists all the 109 projects when public flag is not used and all the 37 public projects when public flag is used.

PR #184, opened 2024-08-25
feat: Added retention command

Althaf66:
Added new retention commands

`harbor retention create` - create retention policy for the project
`harbor retention list` - list retention execution for the project
`harbor retention delete` - delete retention policy for the project

this PR fixes part of #94

PR #364, opened 2025-03-24
Continuing Work from #184: add retention command

rizul2108:
## Description
This PR continues the work started by [@Althaf66](https://github.com/Althaf66) in #184.

## Changes and Fixes:
Incorporating changes suggested by @bupd


rizul2108:
```bash
rizul@rizu:~/OSS/harbor-cli$ ./harbor-cli tag retention list --project-name asdfghj
┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│  ID             Action         Disabled       Params                    Priority       Scope Selectors                    Tag Selectors                      Template             │
│ ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── │
│  0              retain         false          latestPushedK: 2,         0              repository: [{repoMatches   **}]   &{matches {"untagged":true}  **}   latestPushedK        │
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

```bash
rizul@rizu:~/OSS/harbor-cli$ ./harbor-cli tag retention create
INFO[0036] Added Tag Retention Rule                     
```

```bash 
rizul@rizu:~/OSS/harbor-cli$ ./harbor-cli tag retention delete 
INFO[0010] retention rule deleted successfully          
INFO[0010] Retention Policy deleted successfully        
```

all 3 are happening smoothly

rizul2108:
BUG: Inconsistent behavior from Harbor API after deleting a retention policy.
- Created a project `asdfghj`, added a tag retention rule, and deleted it using harbor-cli.  
- After deletion, one of the APIs correctly reflects the deletion by not showing the retention ID in the project's metadata.  
- However, other APIs still return the retention ID of the deleted policy, causing errors when handling projects with deleted retention policies.  
This is error from harbor API side I think.

Vad1mo:
> BUG: Inconsistent behavior from Harbor API after deleting a retention policy.
> 
> * Created a project `asdfghj`, added a tag retention rule, and deleted it using harbor-cli.
> * After deletion, one of the APIs correctly reflects the deletion by not showing the retention ID in the project's metadata.
> * However, other APIs still return the retention ID of the deleted policy, causing errors when handling projects with deleted retention policies.
> 
> This is error from harbor API side I think.

this seems to be a problem with upstream Harbor. I suggest creating an issue there explaining the problem and referencing it here.

rizul2108:
@bupd the login is failing here also like i mentioned I am not able to login on website also 
the login API is not working I think can you please check if it working for you?

bupd (commented):
@rizul2108 do we currently have a workaround for the tag deletion issue: https://github.com/goharbor/harbor/issues/21783

rizul2108:
No currently not should I add that ?
I can add a check for that if retentionpolicy with that ID exists or not

bupd:
Sure go ahead

rizul2108:
[Screencast from 09-04-25 01:22:53 PM IST.webm](https://github.com/user-attachments/assets/eed299c0-01b4-4858-87f0-332b8faeda92)

rizul2108:
@bupd @Vad1mo 
I noticed that the rules within a retention policy do not have a unique ID field assigned — they always default to 0 like u can see above in the screenshot. This makes it difficult to identify and delete a specific rule, as there's no unique identifier to distinguish them.

As a workaround, we're forced to pass the entire rule object between functions to delete a specific rule, rather than using a simple ID. While this approach works, it feels like a poor and error-prone design.

This seems like a design oversight on Harbor's side.
I'll proceed with the current workaround, but I believe this behavior should be reconsidered.

image

Below is the JSON data that is being rendered in the above image.

```json
{
  "algorithm": "or",
  "id": 132,
  "rules": [
    {
      "action": "retain",
      "params": {
        "latestPushedK": 10
      },
      "scope_selectors": {
        "repository": [
          {
            "decoration": "repoMatches",
            "pattern": "**"
          }
        ]
      },
      "tag_selectors": [
        {
          "decoration": "matches",
          "extras": "{\"untagged\":true}",
          "pattern": "**"
        }
      ],
      "template": "latestPushedK"
    },
    {
      "action": "retain",
      "params": {
        "nDaysSinceLastPush": 3
      },
      "scope_selectors": {
        "repository": [
          {
            "decoration": "repoMatches",
            "kind": "doublestar",
            "pattern": "**"
          }
        ]
      },
      "tag_selectors": [
        {
          "decoration": "matches",
          "extras": "{\"untagged\":true}",
          "kind": "doublestar",
          "pattern": "**"
        }
      ],
      "template": "nDaysSinceLastPush"
    }
  ],
  "scope": {
    "level": "project",
    "ref": 1532
  },
  "trigger": {
    "kind": "Schedule",
    "settings": {
      "cron": ""
    }
  }
}
```

NucleoFusion:
Updated and rebased in #564

Issue #365, opened 2025-03-24
artifact view vulnerabilities does not work


dan-m8t:
I am trying to get a list of vulnebarilities of a specific artifact.

For example:

```
harbor artifact list foo/redis-exporter
┌───────────────────────────────────────────────────────────────────────────────────────────┐
│  ID      Artifact Digest       Type          Size          Vulnerabilities  Push Time     │
│ ───────────────────────────────────────────────────────────────────────────────────────── │
│  208278  sha256:61434ded8      IMAGE         37.76MiB      0                518 day ago   │
│  208277  sha256:bc7d64317      IMAGE         50.51MiB      0                518 day ago   │
│  206990  sha256:974b65642      IMAGE         35.23MiB      0                528 day ago   │
└───────────────────────────────────────────────────────────────────────────────────────────┘
```
Lets say I want the CVEs of ID 208278 which has the lag "latest":

```
harbor artifact view foo/redis-exporter/latest
┌───────────────────────────────────────────────────────────────────────────────────────────┐
│  ID      Artifact Digest       Type          Size          Vulnerabilities  Push Time     │
│ ───────────────────────────────────────────────────────────────────────────────────────── │
│                                                                                           │
└───────────────────────────────────────────────────────────────────────────────────────────┘
```

output is completely empty, I am not sure if 'latest' is the correct reference:

```
harbor artifact view foo/redis-exporter/latest -ojson | grep reference
    "references": null,
```

Am I doing something wrong here?

thanks

Vad1mo:
Should  it not be: 

`harbor artifact view foo/redis-exporter:latest -ojson | grep reference
    "references": null,`

dan-m8t:
nope 
```
harbor artifact view foo/redis-exporter:latest
FATA[0000] invalid project/repository/reference format: foo/redis-exporter:latest
```

Help also states this format: `<project>/<repository>/<reference>`

```
Usage:
  harbor artifact view [flags]

Examples:
harbor artifact view <project>/<repository>/<reference>
```

PR #367, opened 2025-03-26
Fixed bug clean

RohanMishra315:
Fixes: #365 
Issue :    The issue was that this function expected exactly three parts (project, repository, and reference) when splitting by /, which was incorrect because the format should be:

<project>/<repository>:<tag>
<project>/<repository>@<digest>
Some repositories contain / in their names, so a stricter split assumption was breaking valid inputs.

Solution:   updated the parsing logic to correctly handle : and @ as delimiters for the reference part


RohanMishra315:
@bupd Looking for review

pkg/utils/utils.go
```diff
@@ -60,11 +60,32 @@ func ParseProjectRepo(projectRepo string) (string, string) {
...
+
+	var repo, ref string
+	if strings.Contains(repoWithRef, ":") {
+		subParts := strings.SplitN(repoWithRef, ":", 2)
+		repo = subParts[0]
+		ref = subParts[1]
+	} else if strings.Contains(repoWithRef, "@") {
+		subParts := strings.SplitN(repoWithRef, "@", 2)
+		repo = subParts[0]
+		ref = subParts[1]
```
bupd, line 82: this doesn't work for nested repositories

Example
demo.goharbor.io/library/nginx-proxy/nginx:latest

bupd: an easy way to do this is by taking the part after the @ as digest and the part before the / as project name and n's in between as repos/ref

bupd (changes requested):
@RohanMishra315 Thanks for your contributions

I have added suggestion for improvements

Thanks

PR #286, opened 2024-12-17
Showing the name in table format for artifact tags list cmd

Althaf66:
This PR fixes #256 

Due to DCO failure, I created PR for the #257 

**Before**
Screenshot from 2024-11-15 00-34-53

**After**
`harbor artifact tags list`
Screenshot from 2024-11-15 00-54-00

And changed to 'project/repo@digest' format

PR #377, opened 2025-04-05
feat: add ID in table output of `artifact tags list` cmd

rizul2108:
Before:
image
After:
image
 
We can close PR #286 after closing this PR.


Vad1mo:
Where is the tag id needed on the input side? 


I am wondering where is the ID needed! If not there is also no need to display it...

rizul2108:
Yeah it is not required in input side. I was adding this feature to account for PR #286 in which ID was added to the table. If it is not required I can close this PR and we can close #286 also then.

PR #378, opened 2025-04-05
feat: Enhance `user list` cmd to fetch all users by default via paginated API calls

rizul2108:
## Description:

This PR updates the `user list` command in the Harbor CLI to support listing **all users** when no `--page-size` is specified (i.e., set to `0`). It mirrors the behavior of the `project list` command.

## 🛠️ Key Changes:

- Added support for repeated paginated API calls when `--page-size` is `0`, with a default of `page-size=100`.
- Ensures consistent behavior with other commands like `project list`.
- Introduced a check in relevant list views to **avoid printing empty tables**.
  - If no entries are found, a clean message like `"No users found`` is printed instead. This is done for all list commands.

## Behavior:

| CLI Input                         | Result                            |
|----------------------------------|-----------------------------------|
| `harbor user list`               | Fetches **all users**             |
| `harbor user list -n 20`         | Fetches **20 users** (1 page)     |
| `harbor user list -n 100 -p 2`   | Fetches page 2 with 100 users     |

### DIFF
--- a/cmd/harbor/root/labels/list.go
+++ b/cmd/harbor/root/labels/list.go
@@ -47,6 +47,10 @@ func ListLabelCommand() *cobra.Command {
 					log.Error(err)
 				}
 			} else {
+				if len(label.Payload) == 0 {
+					log.Info("No labels found")
+					return nil
+				}
 				list.ListLabels(label.Payload)
 			}
 			return nil
--- a/cmd/harbor/root/labels/list.go
+++ b/cmd/harbor/root/labels/list.go
@@ -40,17 +40,17 @@ func ListLabelCommand() *cobra.Command {
 			if err != nil {
 				log.Fatalf("failed to get label list: %v", err)
 			}
-			FormatFlag := viper.GetString("output-format")
-			if FormatFlag != "" {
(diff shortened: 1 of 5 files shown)


rizul2108:
@bupd
Should I implement this paginated `--page-size=0` logic for all `list` commands?  
I'm wondering if it's necessary for commands where it's unlikely that the number of entries will exceed 10 in most use cases.

Vad1mo:
I am not sure what youo mean here with `--page-size=0`

rizul2108:
I have followed the same logic as in #360.

Previously, the default value for `--page-size` was set to `10`.

I’ve now changed the default to `0`, which acts as a **flag to fetch all entries** using paginated API calls. Since users are unlikely to manually pass `--page-size=0`, this approach cleanly distinguishes between:
- default behavior (fetch all), and  
- user-defined pagination.

You can see the key change here for reference:  
[Change](https://github.com/goharbor/harbor-cli/pull/378/files#diff-67e78ee2856c71d11dbe7007eea83fadbd14374d747a1695e1cb4231baf0c1caR88)

bupd (approved):

Vad1mo (approved):

PR #399, opened 2025-04-13
Add a makefile

scraly:
Having a Makefile is better in order to execute some useful tasks, again and again, without knowing all the flags to add in all the commands :).

I added "build" and "lint" targets, it's a start.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ go.work
 dist/
 /dagger.gen.go
 /internal/*
+harbor-cli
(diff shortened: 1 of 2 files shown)


bupd:
No, we are using [dagger](https://dagger.io/) which makes makefiles obsolete. 

You need to install dagger and run `dagger functions` to check the available cmds to execute. we run our entire ci pipelines in dagger too.

Vad1mo:
we fully using dagger

PR #403, opened 2025-04-14
feat: replace random widths with standard custom widths

### LINKED ISSUE #397: Consistent way to handle column width in Table View
We have separate view for every tables.

For example.
https://github.com/goharbor/harbor-cli/blob/1d08f33a84f88e8435f77f1dec86a4b3d9d101dd/pkg/views/registry/view/view.go#L28

https://github.com/goharbor/harbor-cli/blob/1d08f33a84f88e8435f77f1dec86a4b3d9d101dd/pkg/views/artifact/view/view.go#L29

we can clearly see. we don't have a consistent size for overall table. and this might get way out of hand quite easily. if we don't have a consistent column widths each table will look different from every other table. which will worsen the user experience.

I would suggest. using constant values such as xs, small, medium, large, xl.... and so. for table sizes. and not use numerical values.

Thanks

rizul2108:
### Changes Made
- Introduced support for different table column configurations based on size (`xs`, `s`, `m`, `l`, `xl`).
- Defined `sizeMap` to map size labels to specific column configurations using predefined widths: `WidthXS`, `WidthS`, `WidthM`, `WidthL`, and `WidthXL`.
- Ensured each size layout maintains readability and avoids text truncation where possible.

### DIFF
--- a/pkg/views/base/tablelist/model.go
+++ b/pkg/views/base/tablelist/model.go
@@ -21,11 +21,12 @@ import (
 )
 
 const (
-	WidthXS = 5
-	WidthS  = 10
-	WidthM  = 15
-	WidthL  = 20
-	WidthXL = 25
+	WidthXS  = 4
+	WidthS   = 8
+	WidthM   = 12
+	WidthL   = 16
+	WidthXL  = 20
+	WidthXXL = 24
 )
 
 type Model struct {
--- a/pkg/views/base/tablelist/model.go
+++ b/pkg/views/base/tablelist/model.go
@@ -20,6 +20,14 @@ import (
 	"github.com/goharbor/harbor-cli/pkg/views"
 )
 
+const (
+	WidthXS = 5
+	WidthS  = 10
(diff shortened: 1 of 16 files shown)


rizul2108:
@bupd is this way correct ?

pkg/views/base/tablelist/model.go
```diff
@@ -20,6 +20,14 @@ import (
 	"github.com/goharbor/harbor-cli/pkg/views"
 )
 
+const (
+	WidthXS = 5
+	WidthS  = 10
+	WidthM  = 15
+	WidthL  = 20
+	WidthXL = 25
+)
```
bupd, line 29: good but the width should be multiples of 4. so xs becomes 4, small becomes 8 or 12. like that.

its better to follow the tailwind styling here. https://v3.tailwindcss.com/docs/width

bupd (changes requested):
also we should do this to all table view components.

Issue #415, opened 2025-04-15
Container doesn't recognise env variables for config management


meethereum:
I tried setting up the tool as a docker container, and I failed. But when building from source it works perfectly. Possible bug, but not sure. Would be great if anyone verifies, so the bug label can be added.

## System info

```bash
Linux meethereum 6.8.0-57-generic #59~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Mar 19 17:07:41 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```

## Steps to reproduce:

1. Run harbor cli as a container and update your bashrc/zshrc subsequently. You can look [this](https://github.com/goharbor/harbor-cli?tab=readme-ov-file#container) up for exact steps.

2. Define an env variable like this : `export HARBOR_CLI_CONFIG=/home/meeth/.config/harbor-cli/config.yaml`.

3. Confirm if it is registered  like this : `echo $HARBOR_CLI_CONFIG`

4. login using `harbor login` command. You can use demo credentials provided in the docs.

## Expected output 

The credentials should have been added to the path specified in the env variable.

## Issue

The credentials get added to `/root/.config/harbor-cli/config.yaml` which do not match with my env variable, so I am not able to login. However this works alright when building from source.

Vad1mo:
can you post your full command?

meethereum:
> can you post your full command?

yep. To run it as a container, I referred to the docs and ran this command:

```bash
echo "alias harbor='docker run -ti --rm -v \$HOME/.harbor/config.yaml:/root/.harbor/config.yaml registry.goharbor.io/harbor-cli/harbor-cli'" >> ~/.zshrc
source ~/.zshrc # or restart your terminal
```

meethereum:
@qcserestipy yep you are right. I can finally run harbor cli as a container now 🚀.

meethereum:
Yep, I am working on enhancing read me for new-comers here. Should I go ahead with adding this change as well? @qcserestipy

qcserestipy:
@meethereum I think we can close the issue from here. In case you face any related issues feel free to reopen again.

PR #416, opened 2025-04-15
Updates incorrect config mount location and adds default value for HARBOR_CLI_CONFIG in README.md

Darkhood148:
## Related Issue:
https://github.com/goharbor/harbor-cli/issues/415

## Description:
The location of the config has been incorrrectly set as `$HOME/.harbor/config.yaml` while it should be `$HOME/.config/harbor-cli/config.yaml`. Sets a default HARBOR_CLI_CONFIG in .zshrc and use it to alias harbor. This way we can keep track of updating locations of config file.


bupd (approved):
lgtm

Issue #446, opened 2025-04-25
BUG: Undefined behaviour in project create cmd


bupd:
## Summary

When executing project create with argument it automatically creates a private project with default options instead of  asking the user for the necessary things which should be avoided.

## Steps to Reproduce
1. Run `harbor-cli project create [project_name]` you can use any project name
2. See you have not asked for the options. and project been  directly created.

## Expected Behaviour
- `project create cmd` should ask user for the necessary options.
- and then create project with the user defined options

PR #447, opened 2025-04-28
Fix: Ensure `project create cmd` asks for necessary options

Alokzh:
Fixes: #446 

This PR fixes a bug where `project create [project_name]` would automatically create a project without prompting the user for necessary options.

### Changes
- Modified the command handler to always prompt the user, even when a project name is provided as an argument
- Tested the command both with and without a project name argument, ensuring that in both cases the user is prompted for the necessary options.

**Screenshot:**

Screenshot from 2025-04-29 02-32-15


cmd/harbor/root/project/create.go
```diff
@@ -38,17 +38,11 @@ func CreateProjectCommand() *cobra.Command {
...
-
-				if opts.ProxyCache && opts.RegistryID == "" {
-					log.Errorf("Use the --registry-id flag with a registry ID")
-				} else {
-					err = api.CreateProject(opts)
-				}
-			} else {
-				err = createProjectView(createView)
+				createView.ProjectName = args[0]
 			}
```
bupd, line 42: I would suggest checking if the flags were given if no flags were given or any flags missing we should show the TUI otherwise not. 

so.

```
/bin/harbor-dev project create kumar --public=true --proxy-cache=false --registry-id=7 --storage-limit=-1
```
The above cmd should not spawn the TUI

bupd (changes requested):
@Alokzh Thanks for your contributions.

I have left some suggestions for improvements

bupd (changes requested):
@Alokzh Thanks for your contributions.

I have left some suggestions for improvements

cmd/harbor/root/project/create.go
Alokzh: Thnx @bupd I'll improve it as you said

Vad1mo:
> * Modified the command handler to always prompt the user, even when a project name is provided as an argument

harbor-cli should be usable without a TUI, for example in the context of a script..

Alokzh:
@bupd , @Vad1mo Can you please help me with some doubts here 
1. We should pass `--registry-id` only when `--proxy-cache=true` & if it is false we should directly create the project. Am I right here ?
2. In case of `--proxy-cache=true` I am just not able to figure out what our `--registry-id` should be ( In case of TUI it asks to select one options from the prompts directly )

bupd:
> 1. We should pass --registry-id only when --proxy-cache=true & if it is false we should directly create the project. Am I right here?

Yes

> 2. In case of --proxy-cache=true I am just not able to figure out what our --registry-id should be ( In case of TUI it asks to select one options from the prompts directly )

as a user we need to specify the --registry-id

you can get the id by running the below cmd
```
./harbor-cli registry list 
```

which will print out list of registries with their IDs 

Hope this helps. thanks

Alokzh:
Thnx @bupd
Yes this is helpful

Alokzh:
@bupd I have made some changes , please lmk if this fixes the bug correctly or still some changes are needed

cmd/harbor/root/project/create.go
```diff
@@ -42,12 +43,23 @@ func CreateProjectCommand() *cobra.Command {
...
+
+			if opts.ProjectName != "" &&
+				(flagsSet["public"] && flagsSet["storage-limit"] &&
+					(!opts.ProxyCache || (opts.ProxyCache && flagsSet["proxy-cache"] && flagsSet["registry-id"]))) {
 				log.Debug("Attempting to create project using flags...")
 				err = api.CreateProject(opts)
 				ProjectName = opts.ProjectName
 			} else {
-				log.Debug("No project name provided. Switching to interactive view...")
+				log.Debug("Some flags missing. Switching to interactive view...")
```
bupd, line 62: @Alokzh can we remove this we actually don't need to too verbose in logs here.

cmd/harbor/root/project/create.go
```diff
@@ -42,12 +43,23 @@ func CreateProjectCommand() *cobra.Command {
 				return fmt.Errorf("Error: Proxy cache selected but no registry ID provided. Use --registry-id.")
 			}
 
-			if opts.ProjectName != "" {
+			if !opts.ProxyCache && opts.RegistryID != "" {
+				return fmt.Errorf("registry ID should only be provided when proxy-cache is enabled")
+			}
+
+			flagsSet := make(map[string]bool)
```
bupd, line 50: checkout how we handle flags in other commands. such as this one

https://github.com/goharbor/harbor-cli/blob/9265b363534b45135edcdee06f42b11f90e8cca3/cmd/harbor/root/registry/create.go#L58

bupd (changes requested):

cmd/harbor/root/project/create.go
Alokzh: @bupd I have tried another approach can you please check if this is correct way to do it because I could not find in other commands i.e how to check if all the arguments are passed or not
Sorry I am just new & starting off with harbor so I am learning along the way

PR #448, opened 2025-04-29
feat: add `project config` command

### LINKED ISSUE #91: Add the metadata command for managing project metadata

rizul2108:
## Introduces the `project config` command, enabling users to manage Harbor project configurations like in UI
image

## Features added:

- **`harbor project config`**  
  Manage project configuration with support for specifying projects by name or ID.  
  Includes options:  
  - `--id` to use project ID instead of project name  
- **Subcommands:**  
  - `harbor project config list` — List current configuration of a Harbor project  
  - `harbor project config update` — Update project configuration interactively or via flags  
 
image


rizul2108:
@bupd why this test-code check is failing I can't understand
I run the tests avail in test directory locally they are all giving `ok` (PASS) status

bupd:
> ### Continuing work from #92
> ## Fix #91

@rizul2108 Please add a clear PR description of changes made and Screenshots of the working cmd

rizul2108:
I am currently working on this. It is not ready for review yet. It needs major changes so i am implementing those currently

Sorry I forgot to mark it as draft doing it just now

rizul2108:
I'm a bit confused about the implementation of the `add config` command.
image

The URL we are referring to is: https://demo.goharbor.io/harbor/projects/1650/configs

From observing this UI, it seems that the API requests made for updating project configurations are not just targeting metadata (i.e., `PUT` or `GET` requests specifically for metadata), but instead are made on the entire project object. This is because metadata alone does not include information such as the project CVE allowlist.

To modify the CVE allowlist, we would need to make a `PUT` request on the entire project — not just via the metadata APIs.

This raises a question:
**Should we include the option to add or modify the project CVE allowlist under the `config add` command in the CLI?**
Doing so would mean our `config` command will have to interact with the broader project APIs and not limit itself to metadata-specific operations.

Additionally, for the `get config` command, if we want to retrieve the CVE allowlist, we would again need to fetch the full project object, not just the metadata.

So my main confusion is:
**To what extent should we try to replicate the behavior/UI of the Harbor project config section in the CLI?**

Does this implementation approach — where the `config` command also handles CVE allowlist and thus uses full project update APIs — make sense? Should I proceed in this direction?

@Vad1mo @bupd need your insights on this

bupd:
> because metadata alone does not include information such as the project CVE allowlist.

@rizul2108 I suggest we don't need to worry about the cve-allowlist in project config

rizul2108:
There is no option to update more than 1 metadata fields in the metadata API requests 
So I will be using the update project API only to update the metadata

cmd/harbor/root/project/config/list.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/goharbor/harbor-cli/pkg/views/project/config/list"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+func ListProjectConfigCmd() *cobra.Command {
+	var err error
+	var projectNameorID string
+	cmd := &cobra.Command{
+		Use:   "list [NAME|ID]",
```
bupd, line 31: Accept only name as an argument and users can give id via flags.

pkg/api/project_config_handler.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/goharbor/go-client/pkg/sdk/v2.0/models"
+
+	"github.com/goharbor/harbor-cli/pkg/utils"
+	log "github.com/sirupsen/logrus"
+)
+
+func ListConfig(isID bool, projectNameOrID string) (project_metadata.ListProjectMetadatasOK, error) {
+	ctx, client, err := utils.ContextWithClient()
+	if err != nil {
+		return project_metadata.ListProjectMetadatasOK{}, err
```
bupd, line 28: Return nil in case of err

pkg/api/project_config_handler.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	isName := !isID
+	response, err := client.ProjectMetadata.ListProjectMetadatas(ctx, &project_metadata.ListProjectMetadatasParams{ProjectNameOrID: projectNameOrID, XIsResourceName: &isName})
+	if err != nil {
+		return project_metadata.ListProjectMetadatasOK{}, err
+	}
+
+	return *response, nil
```
bupd, line 37: Directly return pointers for efficient memory management.

bupd (changes requested):
@rizul2108 please fix the suggestions

cmd/harbor/root/project/config/cmd.go
```diff
@@ -0,0 +1,32 @@
...
+func ProjectConfigCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "config",
+		Short: "Manage project configuration",
+	}
+	cmd.AddCommand(
+		UpdateProjectConfigCmd(),
+		ListProjectConfigCmd(),
+	)
+	cmd.PersistentFlags().BoolVarP(&isID, "id", "", false, "Use project ID instead of name")
```
rizul2108, line 29: @bupd I gave ID flag as persistent flag here.

cmd/harbor/root/project/config/list.go
rizul2108: I will improve the use description in each command

pkg/api/project_config_handler.go
rizul2108: resolved

pkg/api/project_config_handler.go
rizul2108: resolved

rizul2108:
@bupd resolved errors

bupd (approved):
lgtm

PR #451, opened 2025-05-06
Fix: Manpages Remove Dates

bupd:
## Changes Made

- Remove Dates in Manpages


rizul2108:
@bupd https://github.com/goharbor/harbor-cli/blob/97c20c020984acffdf81b8a827176d5693c81599/doc/man-docs/man_doc.go#L43C1-L49C1

can we add this change also `Harbor` is written as `Habor` and `Manuals` is written as `Mannuals`

PR #452, opened 2025-05-06
WIP: Add support for navigating interactive prompts with 'q' key

### LINKED ISSUE #435: `artifact tags list` cmd should exit on q
## Summary
There are two problems with this command

1. when executing harbor artifact tags list - cmd does not exit with one quit. 
2. There are two error messages produced instead of one.

## Screenshots

Image

## Errors
```shell
❯ ./bin/harbor-dev artifact tags list
ERRO Failed to list tags: [GET /projects/{project_name}/repositories/{repository_name}/artifacts/{reference}/tags][404] listTagsNotFound  &{Errors:[0xc00041c240]}
ERRO failed to list tags: [GET /projects/{project_name}/repositories/{repository_name}/artifacts/{reference}/tags][404] listTagsNotFound  &{Errors:[0xc00041c240]}
```

This error message should be user friendly

qcserestipy:
Fixes #435 

## Problem
Currently, the Harbor CLI's interactive selection prompts don't properly support navigation between different selection levels. When a user presses 'q':

- The UI exits, but the command doesn't properly handle this special value
- Users can't easily go back to previous selection screens
- The selection process can get stuck in loops or require restarting the command

## Solution
This PR implements proper handling of the 'q' key in interactive prompts to enable intuitive navigation:

- At the top-level prompt (project selection), 'q' exits the command
- At lower-level prompts (repository, artifact selection), 'q' goes back to the previous selection level
- Clear feedback messages inform users about navigation options

### DIFF
--- a/cmd/harbor/root/artifact/tags.go
+++ b/cmd/harbor/root/artifact/tags.go
@@ -89,9 +89,12 @@ func ListTagsCmd() *cobra.Command {
 			} else {
 				projectName, err = prompt.GetProjectNameFromUser()
 				if err != nil {
-					log.Errorf("failed to get project name: %v", utils.ParseHarborErrorMsg(err))
+					log.Errorf("failed to get project name: %v", err)
 				}
 				repoName = prompt.GetRepoNameFromUser(projectName)
+				if repoName == "" {
+					return
+				}
 				reference = prompt.GetReferenceFromUser(repoName, projectName)
 			}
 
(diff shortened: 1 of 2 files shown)


qcserestipy:
The fix was easier in the end. To keep it consistent with the rest of the code a simple `return` missing in case `repoName` is empty as selection.

qcserestipy:
Since it was a minor change I am going to merge this immediately.

Issue #469, opened 2025-05-14
CLI info command truncates BannerMessage field


meethereum:
When running `harbor-cli info`, the `BannerMessage` field doesn't display the full message string.
If this is intentional because the message is too huge, still I think the table needs better formatting(since the message itself is not visible). wdyt? 

![Image](https://github.com/user-attachments/assets/59cf3c66-4518-4a1e-8bdc-e2e06f7db04a)

bupd:
Banner message should have only message content and should be truncated we should remove the json. In the banner message

PR #471, opened 2025-05-15
fix: Show Banner messsage in system info instead of json

afeefuddin:
- Fixes: #469

Before: 
image

After:

Screenshot from 2025-05-16 03-13-50

### DIFF
--- a/pkg/views/info/list/view.go
+++ b/pkg/views/info/list/view.go
@@ -14,6 +14,7 @@
 package list
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"reflect"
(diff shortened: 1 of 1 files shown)


pkg/views/info/list/view.go
```diff
@@ -147,6 +148,16 @@ func createRows(data interface{}, rows *[]table.Row) {
 			var value string
 			if fieldName == "TotalStorageConsumption" {
 				value = fmt.Sprintf("%s", utils.FormatSize(int64(field.Interface().(int64))))
+			} else if fieldName == "BannerMessage" {
+				var bannerMessage struct {
+					Message string `json:"message"`
+				}
```
bupd, line 154: declare types on top of the file for better consistency.

bupd (approved):
overall lgtm

@afeefuddin Thanks for your contribution, I have added a suggestion

pkg/views/info/list/view.go
afeefuddin: Done!

Issue #480, opened 2025-05-22
pointer positioning on context change


PrathamX595:
when the context is changed, instead of the star pointer (*) moving to the selected context, the selected context moves to the top, creating confusion about whether the context has changed.

![Image](https://github.com/user-attachments/assets/62b3fc65-ce55-41de-9760-38e2677da413)

---

### Solution

we can make the pointer move when ever the context is changed

PR #481, opened 2025-05-22
feat: fixing the pointer positioning with change in context

PrathamX595:
Fixes: #480 

---

these changes make it so that when the context is changed instead of the context coming to the top, the pointer simply moves to the selected context

---

Before:
Screenshot 2025-05-22 113058

After:
Screenshot 2025-05-22 112510


Vad1mo (approved):

bupd (approved):
lgtm

PR #101, opened 2024-06-02
Add: Project Robot Account Management commands

bupd:
## Fixes #100 

This PR adds system robot account cmds to manage project wide robot accounts in harbor cli.

Commands to implement:
- [x] `harbor project robot list`        -- list robot accounts
- [x] `harbor project robot get`        -- get a robot account
- [x] `harbor project robot delete`   -- delete robot account
- [x] `harbor project robot create`   -- create a robot account 
- [x] `harbor project robot refresh`   -- refresh the secret of robot account
- [x] `harbor project robot update` -- update a robot account

This PR handles project wide robot accounts.

PR #489, opened 2025-06-05
WIP: Finalizing Robots Account cli command

qcserestipy:
# Objective:

- Implement robot account management in CLI which helps to run automated operations.
- Add commands to manage both system and project robot accounts.
- Manage permissions for the robot account 
- Continue work of @bupd in PR #101 

# Tasks

Add CLI Commands for Robot Account Management on System Level:

- [ ] Create: Add a new robot account.
- [ ] Update: Modify an existing robot account.
- [ ] Delete: Remove a robot account.
- [ ] View: Display details of robot accounts.
- [ ] Refresh: Refresh the robot secret.
- [ ] Ensure no sensitive information is leaked during managing robot secrets.
- [ ] Conceptualize and implement a way for permission selection in the CLI

Finalizing work of @bupd on CLI Commands for Robot Account Management on Project Level.

# Rationale

Implementing secure and efficient robot account management in the CLI is crucial for enabling automated operations and replication, enhancing overall project security and functionality.

# Changes

I want to suggest a change in the way how permissions can be selected for the robot account creation. I used a `huh` Form to emulate a grid like experience we also have in the harbor UI. I would like to use it as a discussion basis:

https://github.com/user-attachments/assets/399c75b5-1dd5-4c6c-87c9-a33d6b576dca

### DIFF
--- a/cmd/harbor/root/project/robot/create.go
+++ b/cmd/harbor/root/project/robot/create.go
@@ -1,6 +1,7 @@
 package robot
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/atotto/clipboard"
@@ -15,7 +16,6 @@ import (
 	"github.com/spf13/viper"
 )
 
-// to-do add json file as input and getting json file as output from input.
 func CreateRobotCommand() *cobra.Command {
 	var (
 		opts        create.CreateView
@@ -92,12 +92,16 @@ func CreateRobotCommand() *cobra.Command {
 
 			FormatFlag := viper.GetString("output-format")
 			if FormatFlag != "" {
-				utils.PrintPayloadInJSONFormat(response.Payload)
+				name := response.Payload.Name
+				res, _ := api.GetRobot(response.Payload.ID)
+				utils.SavePayloadJSON(name, res.Payload)
 				return
 			}
 
-			create.CreateRobotSecretView(response.Payload)
+			name, secret := response.Payload.Name, response.Payload.Secret
+			create.CreateRobotSecretView(name, secret)
 			err = clipboard.WriteAll(response.Payload.Secret)
+			fmt.Println("secret copied to clipboard.")
 		},
 	}
 	flags := cmd.Flags()
--- a/cmd/harbor/root/project/robot/create.go
+++ b/cmd/harbor/root/project/robot/create.go
@@ -2,7 +2,6 @@ package robot
 
 import (
 	"fmt"
-	"os"
 
 	"github.com/atotto/clipboard"
 	"github.com/goharbor/go-client/pkg/sdk/v2.0/models"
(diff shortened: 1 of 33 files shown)


qcserestipy:
As discussed with @bupd, it could be also great to have a flag that can read permissions for a robot account from a yaml / json file and then skip the big selection screen.

qcserestipy:
# Updated TableGrid Model
I updated the table grid model to also allow for toggling columns. Columns and rows can be toggled in a vim style with `ctrl+j`/`ctrl+k` for rows and `ctrl+h`/`ctrl+l` for columns.


https://github.com/user-attachments/assets/27504f4d-21f1-4f04-948a-c128f380ea51

bupd (changes requested):
Here is my feedback after testing.

1. Currently I need to press `Ctrl + j` two times to select a row. and same to unselect a row.
2. `Ctrl + j` is the only thing that works now. 

- `ctrl+k` and column selection using `ctrl+h / l` doesn't work
 
3. Wrong Error Message on trying to create a Robot Account with the same name.

```
❯ ./bin/harbor-dev project robot create --project demo --name kumaruu --description power --duration -1
Error: failed to create robot: bad request permission: robot-account:create
```

qcserestipy:
@bupd Concerning point 3.: There was actually still a small bug in the permissions grid, because in the Harbor swagger api the Robots resource is called `robot` and not `robot-account`. I fixed this now. Still when trying to create a robot that already exits we get the following error:

```
Error: failed to create robot: response status code does not match any response statuses defined for this endpoint in the swagger spec (status 409)
```

This is because in v2.10.0 the go-client does not support return code 409 in the swagger definition. Also in v2.13.1 of the go-client it won't be their. Is this a feature request that we have to open in the harbor repository?

bupd:
> This is because in v2.10.0 the go-client does not support return code 409 in the swagger definition. Also in v2.13.1 of the go-client it won't be their. Is this a feature request that we have to open in the harbor repository?

yeah feel free to do so.

To fix this we can add a validation like in the UI to check if the robot account name is already created

```
❯ curl -u 'admin2:Harbor12345' \
  -H "Content-Type: application/json" \
  -H "Accept: application/json" \
  -X GET "https://harbor-oidclab.bupd.xyz/api/v2.0/robots?q=name=zzz"
```

using the above as a validation before actually do the POST request would solve this issue.

Thanks

cmd/harbor/root/project/robot/list.go
bupd, line 1:
```
❯ ./bin/harbor-dev project robot list test -o yamlss
~/code/OSS/harbor-cli-pr robots
❯ ./bin/harbor-dev project list -o jslns
Error: unable to output in the specified 'jslns' format
```
should not allow other formats.

cmd/harbor/root/project/robot/create.go
bupd, line 1: expiration should be a validation field. with default value being -1 and not 0.

bupd: toggle j for on and k for off. in the permission selection.

bupd (approved):
lgtm 

changes requested here would be created in next PR

Issue #230, opened 2024-10-30
The release process is regenerating binaries. 


Vad1mo:
The current release process is using goreleaser to make the binaries, on the other hand when we create the images for the harbor-cli, we are rebuilding the binaries. IMO it would make sense to reuse the binaries instead of regenerating them. 

- same process 
- same signatures and hashes of generated binaries.

bupd:
It seems we can use goreleaser to create docker images too. 
- https://goreleaser.com/customization/docker/
- https://goreleaser.com/customization/sbom/

bupd:
Also we can use goreleaser to publish to AUR and homebrew too.

bupd:
I am brainstorming here.

insteading building binaries inside dagger,  We can update the Release pipeline to use the binaries generated by the goreleaser and package them into containers created by dagger. so we can use the same sbom. too

bupd:
@danish9039 
yes, create binaries and sbom with goreleaser, and publish to github release.

for the dagger part, we should use the binary from goreleaser. and attach the sbom to the container.

danish9039:
@bupd ,
I’ve completed the initial implementation for this PR and would really appreciate your review and any guidance you can share.
Please let me know if there’s anything I should improve or if I’ve missed something. Thanks in advance!

PR #491, opened 2025-06-13
used goreleaser to create binaries and SBOM ,attached it to caintainer

danish9039:
### Related Issues

Fixes 
- https://github.com/goharbor/harbor-cli/issues/230

---

### Description

This PR enhances the CI/CD pipeline by using  **GoReleaser**   to build  binaries and SBOM and use it in **Dagger** containers . 

---

#### Dagger Improvements

* Added SBOM generation for both archives and binaries.
* Updated container build process to use GoReleaser-generated binaries (`publishImage` function in `.dagger/main.go`).
* Included SBOM files in container images at `/usr/share/doc/harbor/`.
* Added OCI labels referencing SBOM locations.

#### GitHub Workflow

* Streamlined the build process to use Dagger consistently.
* Added separate jobs for snapshot and release builds.

---

### New Behavior

* **On tag push**:

  * Run GoReleaser to build binaries, generate SBOM, and publish to GitHub Releases.
  * Use generated binaries and SBOM for container image creation.

* **On main branch push**:

  * Run GoReleaser in snapshot mode (no publishing).
  * Use snapshot binaries and SBOM for container image creation.

---

### Testing

* [x] Verified local builds with `dagger call publish-image`

image

* [x] Tested snapshot builds on `main` branch
* [x] Verified release builds with version tags
* [x] Confirmed SBOM generation and inclusion


danish9039:
The vulnerability check was previously failing due to known security issues in Go standard libraries, as seen in the [pre-update report](https://github.com/user-attachments/files/20736393/vulnerability-check-before-report.txt)  by running `dagger call vulnerability-check-report export --path=vulnerability-check.report`  in my local machine.

Updated Go to version 1.24.4, which resolves the reported vulnerabilities.

Also upgraded Dagger to version 0.18.10 for compatibility.

After the updates, re-running the check shows no remaining vulnerabilities, as confirmed in the [post-update report](https://github.com/user-attachments/files/20736395/vulnerability-check-after-report.txt).

.goreleaser.yaml
```diff
@@ -54,7 +57,7 @@ snapshot:
 
 release:
   name_template: "HarborCLI {{.Tag}}"
-  draft: true                 # Set to false to ensure that releases are published, not kept as drafts
```
bupd: revert this

setting this to false will disable the drafts which will auto publish releases to github. We don't need automatic releases on every goreleaser run.

https://goreleaser.com/customization/release/#github

.github/workflows/default.yaml
```diff
@@ -181,18 +181,21 @@ jobs:
...
 
+      - name: Run Snapshot with GoReleaser
+        uses: dagger/dagger-for-github@v7
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          version: "latest"
+          verb: call
+          args: "snapshot-release"
+
```
bupd: this is not needed in this step. this step runs on every push to main. so we don't want to create new release on every push to main.

.goreleaser.yaml
danish9039: done

.github/workflows/default.yaml
danish9039: addressed this in latest commit

danish9039:
Hi @bupd , sorry for the delay , looks like i should increase the code coverage for ` pkg/utils/utils.go ` in order to pass the CI .

danish9039:
@bupd , can you check this , looks like a lot of files needs their code coverage to be improved ?

bupd:
> @bupd , can you check this , looks like a lot of files needs their code coverage to be improved ?
@danish9039 
It is fine to not have code coverage.
we are discussing on how to move this forward. since we need to solve two issues together. 
1. re-building binaries in dagger.
2. missing sbom & and different 

so the best way would be to use goreleaser to create binaries and put those binaries in respective container images using dagger and push to registry and so you can also attach the sbom created by the goreleaser to the container. since the container is from scratch and has only harbor-cli binary.

bupd:
Please do join the next harbor-cli community meeting so we can discuss this and plan.

Thanks for your contributions

PR #496, opened 2025-06-20
fix: default FormatUrl to use https instead of http

### LINKED ISSUE #495: CLI Login Should use https by default not http
Currently The CLI logins with http and not https

```
❯ ./bin/harbor-dev login demo.goharbor.io/
Login successful for harbor-cli at http://demo.goharbor.io
```

By default CLI should always use https unless http is specified by the user.

devc007:
Previously, FormatUrl prepended http:// to server addresses without a scheme. This change updates the function to use https:// by default, improving security and aligning with best practices for Harbor CLI connections.
Fixes #495


devc007:
Hi @bupd, is this PR good to go for merge, or would you prefer me to make further improvements?

Issue #493, opened 2025-06-19
Pagination for listing repos is ignored


mistersouls:
Pagination does not work. It still responses same response even if adding `--page` or `--page-size`.
The command is: `harbor repo list $project --page "$page" --page-size "$page_size" -o json`

Vad1mo:
@jelly-afk feel free

PR #502, opened 2025-06-23
Fix pagination in repo list

jelly-afk:
fixes: #493 

ListFlags are now properly passed to ListRepository and used for pagination and filtering.
Previously, ListFlags were not being utilized in ListRepository calls.


bupd (approved):
lgtm

Issue #355, opened 2025-03-18
Project List should support query


Vad1mo:
Searching should support the capabilities of Harbor:

> string(query) | Query string to query resources. Supported query patterns are "exact match(k=v)", "fuzzy match(k=~v)", "range(k=[min~max])", "list with union releationship(k={v1 v2 v3})" and "list with intersetion relationship(k=(v1 v2 v3))". The value of range and list can be string(enclosed by " or '), integer or time(in format "2020-04-09 02:36:00"). All of these query patterns should be put in the query string "q=xxx" and splitted by ",". e.g. q=k1=v1,k2=~v2,k3=[min~max]
> -- | --
>

rizul2108:
@bupd The `query` flag is currently in the `project list` command, but the logic for handling it has not been implemented yet. Additionally, it is not included in the `project search` command. Would it be appropriate to move the `query` flag to the `project search` command and implement the logic there?

rizul2108:
query flag should be added to project search or list as currently in harbor api it is in list command parameters not in search

rizul2108:
@bupd

bupd:
@NucleoFusion please do share your plan and your overall understanding in the issue before creating a PR

NucleoFusion:
@bupd Okay let me have a detailed look on the issue and code, and think for a little. Then Ill get back to you

bupd:
> Also, I wanted to know if there is a docs page etc for the goharbor/go-client repo? I had to manually look at the go code in the repo, and figure out the stuff with the code comments because I couldnt find a docs page.

there is no docs for the harbor-client
except the swagger file

NucleoFusion:
This has been resolved in #533 .

PR #533, opened 2025-08-18
enhancement: Support for querying in project list command

NucleoFusion:
Adding support for 3 flags
- `--fuzzy`, expects a key=value string. Fuzzy matches the value to the key for output.
- `--match`, expects a key=value string. Exactly matches the value to the key for output.
- `--range`, expects a key=min~max string. matches the range to the key for output.

Solves #355 

https://github.com/user-attachments/assets/ec5fbeb3-2186-4855-a794-dbbe425dc4fb

### DIFF
--- a/doc/cli-docs/harbor-project-list.md
+++ b/doc/cli-docs/harbor-project-list.md
@@ -15,13 +15,15 @@ harbor project list [flags]
 ### Options
 
 ```sh
+      --fuzzy strings   Fuzzy match filter (key=value)
   -h, --help            help for list
+      --match strings   exact match filter (key=value)
       --name string     Name of the project
       --page int        Page number (default 1)
       --page-size int   Size of per page (0 to fetch all)
       --private         Show only private projects
       --public          Show only public projects
-  -q, --query string    Query string to query resources
+      --range strings   range filter (key=min~max)
       --sort string     Sort the resource list in ascending or descending order
 ```
 
--- a/doc/man-docs/man1/harbor-project-list.1
+++ b/doc/man-docs/man1/harbor-project-list.1
@@ -14,8 +14,16 @@ List projects
 
 
 .SH OPTIONS
+\fB--fuzzy\fP=[]
+	Fuzzy match filter (key=value)
+
+.PP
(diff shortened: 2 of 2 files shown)


cmd/harbor/root/project/list.go
```diff
@@ -142,3 +161,72 @@ func fetchProjects(listFunc func(...api.ListFlags) (project.ListProjectsOK, erro
...
+
+		parts = append(parts, fmt.Sprintf("%s=[%s~%s]", kv[0], rng[0], rng[1]))
+	}
+
+	return strings.Join(parts, ","), nil
+}
+
+// Validates Key provided by user for ListFlags.Q
+func validKey(key string) error {
+	keys := []string{"name", "project_id", "public", "creation_time"}
```
bupd: I would suggest to add repo_count and owner info to valid keys
```suggestion
	keys := []string{"name", "owner_name", "owner_id", "project_id", "repo_count", "public", "creation_time"}
```

NucleoFusion: Yep I tried it, but as far as I researched those are not supported, or have some different key they go by
Like projectName goes by name Access Level by public (bool)
Do you have somewhere I could find this? I searched on google, and referred to the go sdk code for all docs.

NucleoFusion: I will try to check if `owner_name` or `owner_id` works, but `repo_count` did not work for me.

NucleoFusion: so `owner_name` doesnt work, and neither does `repo_count` (at least for me, I just tried to give an abnormal value, `'somename'` and `999`) to get a nil output, but it returns normal output

bupd:
> Hmm I cant seem to find any guidelines regarding testing, and the test/ directory seems to be empty too Am I supposed to add testing? If so, could you please tell me how?

currently we have only e2e tests and no unit tests. but you are free to write mock tests.
before that we need to remove the dependency on demo.goharbor.io for running e2e tests.

PTAL https://github.com/goharbor/harbor-cli/issues/538

bupd:
@NucleoFusion lets track the test in issue: #538

PR #85, opened 2024-05-28
Add Project Member Management Commands

bupd:
## Overview:
The current CLI lacks the ability to **manage project members**, a feature already present in the WebUI.
Adding a `member` command will significantly enhance the CLI's functionality, making it a powerful tool for handling project members directly from the command line. 

This addition will facilitate easier and more efficient management of project members, aligning the CLI's capabilities with those of the WebUI.

### New Commands:
The member command will include the following subcommands:
- list: List all members in the project.
  -  Usage: `harbor member list <project_name_or_id>`
  - Description: Retrieves and displays a list of all members in the specified project.
  
- create: Create a new member in the project.
  - Usage: `harbor member create <project_name_or_id>`
  - Description: Adds a new member to the specified project using the provided member information.

- delete: Delete a member from the project.
  - Usage: `harbor member delete <project_name_or_id> <member_id>`
  - Description: Removes specified members from the project.
 
- view: View details of a specific member.
  - Usage: `harbor member view <project_name_or_id> <member_id>`
  - Description: Retrieves and displays detailed information about a specific project member.

- update: Update an existing member in the project.
  - Usage: `harbor member update <project_name_or_id> <member_id>`
  - Description: Updates the details of an existing project member with the provided information.

This PR adds the commands to manage project members.

PR #540, opened 2025-09-03
[WIP] Updating the Project Member Commands

NucleoFusion:
**Updating and fixing #85 's progress on Project Member Commands.**

#### Major

- [ ] Adding fuzzy search for the project selector
- [ ] The update bubbletea view is broken (I think two of the models are being shown at once)
- [ ] For creation, some fields such as GroupID and UserID are difficult to for a user to remember, so maybe move to using names for them, with a selection list like it is for project.
- [ ] Also in these fields, an (optional) / required label should be there
- [ ] Mismatch in conventions, list/update/view have projectName as a subcommand variable whereas create/delete uses a flag `--project`.
- [ ] View command and the Update command, imo, should work with username instead of id, and maybe `--id` as a flag.

#### Minor

- [ ] The bubbletea instance uses full screen, which I think should only use as much as required (makes it look more like a TUI instead of CLI)
- [ ] Convert the forms into a step-by-step form instead of an all at once. (Kinda like each step is its seperate view?).


pkg/api/member_handler.go
```diff
@@ -0,0 +1,214 @@
...
+
+// List Members in project
+func ListMembers(projectName string) (*member.ListProjectMembersOK, error) {
+	ctx, client, err := utils.ContextWithClient()
+	if err != nil {
+		return nil, err
+	}
+	response, err := client.Member.ListProjectMembers(
+		ctx,
+		&member.ListProjectMembersParams{ProjectNameOrID: projectName},
```
bupd: always set `X-Is-Resource-Name` to true in the params. so project will only be got by the name and not by id.

cmd/harbor/root/project/member/list.go
```diff
@@ -0,0 +1,80 @@
...
+				list.ListMembers(members.Payload, false)
+			}
+			return nil
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.Int64VarP(&opts.Page, "page", "", 1, "Page number")
+	flags.Int64VarP(&opts.PageSize, "page-size", "", 10, "Size of per page")
+	flags.StringVarP(&opts.EntityName, "name", "n", "", "Member Name to search")
```
bupd: add `project-id` flag

cmd/harbor/root/project/member/list.go
```diff
@@ -0,0 +1,80 @@
...
+					return fmt.Errorf("failed to get project name: %v", err)
+				}
+			}
+
+			members, err := api.ListMember(opts)
+			if err != nil {
+				return fmt.Errorf("failed to get members list: %v", err)
+			}
+
+			FormatFlag := viper.GetString("output-format")
```
bupd, line 78: if the output format is neither json nor yaml. fail the cmd with invalid format flag error

cmd/harbor/root/project/member/view.go
```diff
@@ -0,0 +1,82 @@
...
+		Example: "  harbor project member view my-project [memberID]",
+		Args:    cobra.MaximumNArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var err error
+			if len(args) == 1 {
+				opts.ProjectNameOrID = args[0]
+			} else if len(args) == 2 {
+				opts.ID, _ = strconv.ParseInt(args[1], 0, 64)
+				opts.ProjectNameOrID = args[0]
+			} else if opts.ProjectNameOrID == "" || opts.ID == 0 {
```
bupd: same here. do not use `projectnameorid` 
- use `projectname` and `projectid` -- project-id should be a flag
- projectname should be an argument

cmd/harbor/root/project/member/view.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+// NewGetRegistryCommand creates a new `harbor get registry` command
+func ViewMemberCommand() *cobra.Command {
+	var opts api.GetMemberOptions
+
+	cmd := &cobra.Command{
+		Use:     "view [ProjectName Or ID] [member ID]",
```
bupd:
```suggestion
		Use:     "view [MemberName]",
```

change the above to only have projectname as an argument

& add `project-id` & `project-name` as a flag.

cmd/harbor/root/project/member/view.go
```diff
@@ -0,0 +1,82 @@
...
+		Short:   "get project member by ID or name",
+		Long:    "get member details by MemberID",
+		Example: "  harbor project member view my-project [memberID]",
```
bupd:
```suggestion
		Example: "  harbor project member view my-project",
```

cmd/harbor/root/project/member/view.go
bupd: from the api perspective. 

1. i would give a member name (eg kumar)
2. If user did not give the project name api will fetch all the projects that the member is present. if he is an admin it will show all projects.
3. and for the selected project show the details and the permission.

cmd/harbor/root/project/member/update.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/spf13/cobra"
+)
+
+// NewGetRegistryCommand creates a new `harbor get registry` command
+func UpdateMemberCommand() *cobra.Command {
+	var opts api.UpdateMemberOptions
+	var roleID int64
+
+	cmd := &cobra.Command{
+		Use:     "update [ProjectName Or ID] [member ID]",
```
bupd: same here just stick with member name

cmd/harbor/root/project/member/update.go
bupd, line 1: `project member update` should fail if the project has no members in it.

bupd: if no members found give the error `No members found in project`

cmd/harbor/root/project/member/delete.go
bupd, line 1: x-is-resource-name - set this to true in all cmds and just use names instead of id for querying

cmd/harbor/root/project/member/create.go
bupd, line 1: do not ask for group name. if the authentication mode is not ldap.

bupd (changes requested):
@NucleoFusion Thanks for your contributions

I have added some suggestions for improvements

bupd (approved):
lgtm

PR #541, opened 2025-09-05
Doc change

neel-1414:
Go language version in Contribution.md changed from >=1.20 to >=1.24.


neel-1414:
Thanks for the feedback @bupd !
I've rebased the branch onto the latest main and removed the redundant commits, so now the PR contains only the intended doc change.

PR #547, opened 2025-09-13
Feature: Moving to pure Dagger based CI/CD Pipeline from Goreleaser

NucleoFusion:
## Completed Release

The release binaries are here,
https://github.com/NucleoFusion/harbor-cli/releases/tag/v0.2.54

image

We only miss the SBOM's from the current release structure.

### Apt repo

The apt repo files are here,
https://github.com/NucleoFusion/harbor-cli/tree/gh-pages

PS: Now updated so that only `Packages.gz` is there

### Publish and Image

The images are at,
https://hub.docker.com/repository/docker/nucleofusion/harbor-cli

PS: Now re-use binaries instead of re-creation

### Homebrew

The homebrew formula/PR for cron is at,
https://github.com/goharbor/homebrew-tap/pull/5

PS: Now supports mac and linux both

## Tracking Issue 
https://github.com/goharbor/harbor-cli/issues/553

### Releases

- [x] Build binaries - linux, windows, mac, bsd - amd64, arm64, armv6
- [x] Archives (tar.gz) w/ Checksum
- [x] .deb, .rpm w/ nFPM
- [x] Release Publish
- [x] Brew Publish

### Misc

- [x] Doc Gen
- [x] Code Coverage
- [x] Lint & Lint Report
- [x] Tests
  - [x] Run Tests
  - [x] Test Coverage
  - [x] Test Report

### DIFF
--- a/.github/workflows/default.yaml
+++ b/.github/workflows/default.yaml
@@ -66,13 +66,6 @@ jobs:
               exit 1
           fi
 
-      # - uses: reviewdog/action-setup@v1
-      # - name: Run Reviewdog
-      #   env:
-      #     REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      #   run: |
-      #     reviewdog -f=sarif -name="Golang Linter Report" -reporter=github-check -filter-mode nofilter -fail-level any -tee < golangci-lint-report.sarif
-
   vulnerability-check:
     runs-on: ubuntu-latest
     steps:
--- a/.dagger/release.go
+++ b/.dagger/release.go
@@ -25,7 +25,7 @@ func (m *HarborCli) PublishRelease(ctx context.Context,
 		return "", err
 	}
 
-	cmd := []string{"gh", "release", "upload", m.AppVersion}
+	cmd := []string{"gh", "release", "upload", "v" + m.AppVersion}
 	cmd = append(cmd, bins...)
 	cmd = append(cmd, "/dist/checksums.txt")
 	cmd = append(cmd, "--clobber")
@@ -45,7 +45,7 @@ func (m *HarborCli) PublishRelease(ctx context.Context,
 	return ctr.
 		WithWorkdir("/src").
 		// Creating Release
-		WithExec([]string{"gh", "release", "create", m.AppVersion, "--generate-notes"}).
+		WithExec([]string{"gh", "release", "create", "v" + m.AppVersion, "--generate-notes"}).
 		WithExec(cmd).
 		Stdout(ctx)
 }
--- a/.dagger/vuln.go
+++ b/.dagger/vuln.go
@@ -37,9 +37,12 @@ func (m *HarborCli) VulnerabilityCheckReport(ctx context.Context, source *dagger
 	if err != nil {
 		return nil, err
 	}
+
 	report := "vulnerability-check.report"
+	cmd := fmt.Sprintf("govulncheck ./... > %s || true", report)
+
 	return m.vulnerabilityCheck(ctx).
 		WithExec([]string{
-			"sh", "-c", fmt.Sprintf("govulncheck ./... > %s", report),
+			"sh", "-c", cmd,
 		}).File(report), nil
 }


dagger.json
bupd, line 1: why remove the cosign dependency ??

bupd (changes requested):

dagger.json
NucleoFusion: Ohh sorry,
I didn't even realize I deleted that. I will add it back asap.
I was trying to split the code. So had to refactor/remove a lot of code and by mistake forgot.

.dagger/pipeline/brew.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"dagger/harbor-cli/internal/dagger"
+)
+
+func (s *Pipeline) BrewFormula(ctx context.Context, dist *dagger.Directory) (*dagger.Directory, error) {
```
bupd: do we need this since we have this on the brew-tap repo.??

.dagger/README.md
```diff
@@ -1,101 +0,0 @@
-# 🛠️ Harbor CLI Dagger Pipeline
-
-We use [Dagger](https://dagger.io) to define a CI/CD pipeline for building, linting, and publishing the [Harbor CLI](https://github.com/goharbor/harbor-cli).
```
bupd, line 3: do not delete this Readme

dagger.json
```diff
@@ -1,15 +1,8 @@
...
   "sdk": {
     "source": "go"
   },
-  "dependencies": [
-    {
-      "name": "cosign",
-      "source": "github.com/scottames/daggerverse/cosign@8359122a7b90f2c8c6f3165570fdcbec6e923023",
-      "pin": "8359122a7b90f2c8c6f3165570fdcbec6e923023"
-    }
-  ],
```
bupd, line 13: do not delete the cosign dependency... I suggest to add this back. without this signing images would not work as expected

.github/workflows/default.yaml
```diff
@@ -99,23 +102,24 @@ jobs:
...
-      - name: Dagger Version
-        uses: sagikazarmark/dagger-version-action@v0.0.1
-
-      - name: Test Release
-        uses: dagger/dagger-for-github@v7
-        with:
-          version: ${{ steps.dagger_version.outputs.version }}
-          verb: call
-          args: snapshot-release
+  # TODO: Add Dry-run support
```
bupd: do create an issue for this. so we can track this

bupd (changes requested):
I have commented improvements

.github/workflows/default.yaml
NucleoFusion: Added to the existing tracker

.dagger/pipeline/brew.go
```diff
@@ -0,0 +1,72 @@
...
+    system "#{bin}/harbor-cli", "--version"
+  end
+end
+`, appVer,
+		appVer, appVer, shaMap["amd64"],
+		appVer, appVer, shaMap["arm64"],
+	)
+
+	return content
+}
```
bupd:
```suggestion
```

remove this its not needed

.dagger/pipeline/checksum.go
```diff
@@ -0,0 +1,43 @@
...
+		split := strings.Split(v, "/")
+		sums[out] = split[len(split)-1] // Taking only filename
+	}
+
+	content := ""
+	for k, v := range sums {
+		content += fmt.Sprintf("%s %s\n", strings.TrimSuffix(k, "\n"), v)
+	}
+
+	dist = dist.WithFile("checksum.txt", s.dag.File("checksum.txt", content))
```
bupd:
```suggestion
	dist = dist.WithFile("checksums.txt", s.dag.File("checksum.txt", content))
```

.dagger/pipeline/release.go
```diff
@@ -0,0 +1,39 @@
...
+
+func (s *Pipeline) PublishRelease(ctx context.Context, dist *dagger.Directory, token *dagger.Secret) (string, error) {
+	bins, err := DistBinaries(ctx, s.dag, dist)
+	if err != nil {
+		return "", err
+	}
+
+	cmd := []string{"gh", "release", "upload", s.appVersion}
+	cmd = append(cmd, bins...)
+	cmd = append(cmd, "/dist/checksum.txt")
```
bupd:
```suggestion
	cmd = append(cmd, "/dist/checksums.txt")
```

.dagger/README.md
```diff
@@ -1,101 +1,98 @@
...
 
-### 🔧 `BuildDev(platform)`
+### Run Dagger in Development Mode
 
-Builds a development binary for your target platform.
+To start the Dagger session and enable live code reloads:
 
 ```bash
-dagger call build-dev --platform="linux/amd64" export --path=bin/harbor-dev
+dagger develop
```
bupd, line 26:
```suggestion
dagger call build-dev --platform="linux/amd64" export --path=bin/harbor-dev
```

please include this.. back

.github/workflows/default.yaml
```diff
@@ -164,42 +168,42 @@ jobs:
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
-      - name: Build Binary
-        uses: dagger/dagger-for-github@v7
-        with:
-          version: ${{ steps.dagger_version.outputs.version }}
-          verb: call
-          args: build-dev --platform linux/amd64 export --path=./harbor-dev
```
bupd, line 172: add this one back.

.github/workflows/default.yaml
Vad1mo, line 1: I don't understand what the gh action pipeline is rewritten and  many steps are commented out.
What is the intention of the change?

Vad1mo (changes requested):

.github/workflows/default.yaml
NucleoFusion: I beleive that the docker publish and the dry-run/test-run is currently commented out.

.github/workflows/default.yaml
```diff
@@ -242,4 +242,16 @@ jobs:
...
+        if: github.event_name == 'push' && (startsWith(github.ref, 'refs/tags/'))
+        uses: ./.github/actions/publish-and-sign
+        with:
+          REGISTRY_ADDRESS: ${{ vars.REGISTRY_ADDRESS }}
+          REGISTRY_USERNAME: ${{ vars.REGISTRY_USERNAME }}
+        env:
+          GITHUB_TOKEN: ${{ env.GITHUB_TOKEN }}
+          REGISTRY_PASSWORD: ${{ secrets.REGISTRY_PASSWORD }}
+          ACTIONS_ID_TOKEN_REQUEST_URL: ${{ env.ACTIONS_ID_TOKEN_REQUEST_URL }}
+          ACTIONS_ID_TOKEN_REQUEST_TOKEN: ${{ env.ACTIONS_ID_TOKEN_REQUEST_TOKEN }}
```
bupd: publish image should be a separate one and please don't change that. we are not going to push images to dockerhub rather to registry.goharbor.io and it was already implemented. 

I would suggest you to not delete or change the publish-image workflow.

.github/workflows/default.yaml
```diff
@@ -242,4 +242,16 @@ jobs:
           version: "latest"
           verb: call
           # Using a single large line since, dagger parsing with folded/piped command fails due to improper formatting error
-          args: pipeline --source=. --github-token=env://GITHUB_TOKEN --registry-password=env://REGISTRY_PASSWORD --registry-username=${{ vars.REGISTRY_USERNAME }} --registry-addr=${{ vars.REGISTRY_ADDRESS }}
+          args: pipeline --source=. --github-token=env://GITHUB_TOKEN 
+  
+      - name: Publish and Sign Snapshot Image
```
bupd: This is wrong place to put the publish and sign image.

Publish image should be a separate workflow which runs and publishes latest images on every push to main. 

whereas Publish Release should release the binaries to the github, homebrew, and versioned images to the registry.

.github/actions/publish-and-sign/action.yaml
```diff
@@ -24,31 +15,30 @@ runs:
...
         verb: call
-        args: "publish-image-and-sign \
-          --registry='${{ env.REGISTRY_ADDRESS }}' \
-          --registry-username='${{ env.REGISTRY_USERNAME }}' \
-          --registry-password=env:REGISTRY_PASSWORD \
-          --image-tags='${{ env.IMAGE_TAGS}}' \
-          --github-token=env:GITHUB_TOKEN \
-          --actions-id-token-request-url=$ACTIONS_ID_TOKEN_REQUEST_URL \
-          --actions-id-token-request-token=env:ACTIONS_ID_TOKEN_REQUEST_TOKEN"
+        args: "publish-and-sign-image --source=. --registry-address=${{ inputs.REGISTRY_ADDRESS }} --registry-username=${{ inputs.REGISTRY_USERNAME }} --registry-password=env://REGISTRY_PASSWORD --github-token=env://GITHUB_TOKEN --actions-idtoken-request-url=env://ACTIONS_ID_TOKEN_REQUEST_URL --actions-idtoken-request-token=env://ACTIONS_ID_TOKEN_REQUEST_TOKEN"
```
bupd: I suggest to revert this change.

.github/actions/publish-and-sign/action.yaml
```diff
@@ -24,31 +15,30 @@ runs:
...
+    - name: Debug OIDC Environment
       shell: bash
-      env:
-        GITHUB_TOKEN: ${{ inputs.GITHUB_TOKEN }}
-      run: cosign env
+      run: |
+        echo $REGISTRY_PASSWORD
+        echo "ACTIONS_ID_TOKEN_REQUEST_URL: $ACTIONS_ID_TOKEN_REQUEST_URL"
+        echo "ACTIONS_ID_TOKEN_REQUEST_TOKEN present: $([[ -n \"$ACTIONS_ID_TOKEN_REQUEST_TOKEN\" ]] && echo yes || echo no)"
+        cosign version
```
bupd: why this change ??

.github/actions/publish-and-sign/action.yaml
```diff
@@ -24,31 +15,30 @@ runs:
...
       uses: sigstore/cosign-installer@v3.7.0
 
-    - name: Check Env Variables
+    - name: Debug OIDC Environment
       shell: bash
-      env:
-        GITHUB_TOKEN: ${{ inputs.GITHUB_TOKEN }}
-      run: cosign env
+      run: |
+        echo $REGISTRY_PASSWORD
```
bupd: revert this why are echoing registry password. This is unsafe

.github/actions/publish-and-sign/action.yaml
```diff
@@ -2,20 +2,11 @@ name: Publish and Sign Snapshot Image
...
 inputs:
-  IMAGE_TAGS:
-    description: 'Tags for the image, e.g. "latest, v1.0.0"'
-    required: true
-  GITHUB_TOKEN:
-    description: 'GitHub token'
-    required: true
-  REGISTRY_PASSWORD:
-    description: 'Registry password'
-    required: true
```
bupd, line 13: why remove these ??

bupd: these are out of scope for this PR. This PR should only focus on replacing the goreleaser with dagger and should not redo the dagger functions.

.dagger/main.go
```diff
@@ -102,6 +127,10 @@ func (m *HarborCli) init(ctx context.Context, source *dagger.Directory) error {
 		m.GoVersion = match[1]
 	}
 
+	if m.GoVersion == "" {
+		m.GoVersion = "latest"
+	}
```
bupd: we can directly get the go version from go.mod

revert this.

.dagger/main.go
```diff
@@ -63,20 +63,45 @@ func (m *HarborCli) Pipeline(ctx context.Context, source *dagger.Directory,
...
+
+	// Building Binaries
+	dist, err = pipe.Build(ctx, dist)
 	if err != nil {
 		return nil, err
 	}
 
-	return dist, err
+	// Publishing Image
+	imageAddr := pipe.PublishImage(ctx, dist)
```
bupd: I would suggest to move the publish image workflow out of the pipeline. and only keep the pipeline to replace the goreleaser.

.dagger/main.go
```diff
@@ -63,20 +63,45 @@ func (m *HarborCli) Pipeline(ctx context.Context, source *dagger.Directory,
...
+
+	_, err = pipe.Sign(ctx, imageAddr[0])
```
bupd: same here

NucleoFusion:
@bupd my mistake sorry, I was so focused on making it fit with the theme/structure I had, I didnt pay attention that it got out of scope. I will fix this.
(The echo cmd was left there from when I was debugging and forgot to remove, my bad)

.github/workflows/default.yaml
```diff
@@ -99,23 +102,24 @@ jobs:
...
+  #       with:
+  #         fetch-depth: 0
+  #
+  #     - name: Dagger Version
+  #       uses: sagikazarmark/dagger-version-action@v0.0.1
+  #
+  #     - name: Test Release
+  #       uses: dagger/dagger-for-github@v7
+  #       with:
+  #         version: ${{ steps.dagger_version.outputs.version }}
```
Vad1mo: no need to comment out, If we don't need it remove it.

.github/workflows/default.yaml
```diff
@@ -224,16 +228,6 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Push images
```
Vad1mo, line 227: Wher is it?

NucleoFusion: Oh
This is deleted because the publish was being called 2 times previously.

[here it would call the first time](https://github.com/goharbor/harbor-cli/blob/7bd83eb2f369d10381ef95876ebcaeca78e5b980/.github/workflows/default.yaml#L227)

[here the second time](https://github.com/goharbor/harbor-cli/blob/7bd83eb2f369d10381ef95876ebcaeca78e5b980/.github/workflows/default.yaml#L247)

I kept the one after the `publish-release` so that if a build/publish fails it wouldnt push to registry

.dagger/apt.go
```diff
@@ -0,0 +1,102 @@
...
+// /
+// ├── dists/
+// │   └── stable/
+// │       ├── Release
+// │       └── main/
+// │           ├── binary-amd64/
+// │           │   ├── Packages
+// │           │   └── Packages.gz
+// │           └── binary-arm64/
+// │               ├── Packages
```
bupd: Currently we have two of the same binary one with .gz and one without. 

I would suggest to have it only with the .gz

.dagger/apt.go
```diff
@@ -0,0 +1,102 @@
...
+
+// GH-PAGES Structure
+//
+// /
+// ├── dists/
+// │   └── stable/
+// │       ├── Release
+// │       └── main/
+// │           ├── binary-amd64/
+// │           │   ├── Packages
```
bupd: same here remove it.

.dagger/buildDev.go
```diff
@@ -0,0 +1,58 @@
...
+	"strings"
+	"time"
+
+	"dagger/harbor-cli/internal/dagger"
+	"dagger/harbor-cli/utils"
+)
+
+// Create build of Harbor CLI for local testing and development
+func (m *HarborCli) BuildDev(ctx context.Context, source *dagger.Directory, platform string) *dagger.File {
+	err := m.init(ctx, source)
```
bupd:
```suggestion
func (m *HarborCli) BuildDev(ctx context.Context, platform string) *dagger.File {
	err := m.init(ctx, m.Source)
```

.dagger/main.go
```diff
@@ -1,479 +1,98 @@
...
-)
-
-func New(
-	// Local or remote directory with source code, defaults to "./"
-	// +optional
-	// +defaultPath="./"
-	source *dagger.Directory,
-) *HarborCli {
-	return &HarborCli{Source: source}
-}
```
bupd, line 39: add thsi back

.dagger/main.go
```diff
@@ -1,479 +1,98 @@
...
-
 type HarborCli struct {
-	// Local or remote directory with source code, defaults to "./"
-	Source *dagger.Directory
+	Source     *dagger.Directory // Source Directory where code resides
+	AppVersion string            // Current Version of the app, acquired from git tags
+	GoVersion  string            // Go Version used in the current release, acquired from the go.mod file
```
bupd: this too. it was already perfect. no need for changes here.

NucleoFusion:
> Screenshot_2025-11-25_19-14-16
> 
> we still need these two for doing a right release for harbor-cli. and this should be atleast on par with the current releases.

We do have Published and Signed Image.
Let me update the issue as well

.dagger/main.go
NucleoFusion: previously we were using constants to represent this
and since we are taking them from files such as `go.mod` or `git`

.dagger/main.go
NucleoFusion: I have this as the `init()` function now
And it also does more stuff than this, since I have removed dependency on `constants` and moved towards having dynamic values for `GO_VERSION`, `APP_VERSION` and more.

NucleoFusion:
> also looks like we are missing the sbom.json which was already present in the previous releases.
> 
> we still need to add sbom and signing for binaries and publish the same binaries for publish and release.
> 
> so binary downloaded from github and binary in docker image should not have different binaries and different timestamps. currently there is some inconsistencies as i tested this.
> 
> Please do a release on your forked repo and also publish the docker images. through the pipelines and add those in this PR Description. so we can better review and merge this.
> 
> also i believe we have less artifacts in your release as compared to the current releases we have which should not be the case.
> 
> Thanks

Thanks for the review!
To answer some of your queries,

1. The current release has .sbom which makes it so that the dagger release has less release artifacts even though it has more publishes and binaries. ( I will look into and add SBOM, just wanted to clarify)

2. The reason that the docker image and the release have different binaries is because the publish image is not in the CI/CD `"pipeline"` in the dagger. This was done because we did not want to go out of scope of the PR and thus the addition of docker image publish was standalone and integrated with the dagger pipeline through Github Actions and not Dagger Chaining itself.

3. I have added support for release notes successfully (through the `gh` cli itself and no third party tool) please let me know if it is enough for now, or should I look into something that I can possibly customize for harbor-cli itself.

.dagger/apt.go
NucleoFusion: I think this is just how APT wants it? I will check this.

bupd: @NucleoFusion I dont think so. This is an issue. https://github.com/marketplace/actions/cache-apt-packages

.dagger/main.go
bupd: okay then.

PR #563, opened 2025-10-22
fix(robot_delete): adapt robot api handler for deletion by name; adapt robot delete commands for project and system to allow for deletion by name

### LINKED ISSUE #555: ID column in robot list table not present
Cannot see robot ID in `harbor robot list` 

Image

Deleting a robot requires robot ID.

Image

qcserestipy:
Closes #555

## Overview

This PR introduces changes to improve user experience for project and system robot deletion. It introduces the possibility to delete both kinds of robots by name instead of ID.

## Behavior

https://github.com/user-attachments/assets/7d32f1e0-c2ec-4909-857a-3ad64281ea1b

### DIFF
--- a/cmd/harbor/root/project/robot/delete.go
+++ b/cmd/harbor/root/project/robot/delete.go
@@ -78,7 +78,7 @@ Examples:
 				if err != nil {
 					fmt.Print(utils.ParseHarborErrorMsg(err))
 				}
-				robotID = robot.Payload.ID
+				robotID = robot.ID
 			} else if ProjectName != "" {
 				project, err := api.GetProject(ProjectName, false)
 				if err != nil {
--- a/pkg/api/robot_handler.go
+++ b/pkg/api/robot_handler.go
@@ -62,22 +62,19 @@ func GetRobot(robotID int64) (*robot.GetRobotByIDOK, error) {
 	return response, nil
 }
 
-func GetRobotByName(targetRobotName string, projectName ...string) (*robot.GetRobotByIDOK, error) {
+func GetRobotByName(targetRobotName string, projectName ...string) (*models.Robot, error) {
 	if len(projectName) > 1 {
 		return nil, fmt.Errorf("only one project name can be provided")
 	}
 	var listResponse *robot.ListRobotOK
 	var err error
-	ctx, client, err := utils.ContextWithClient()
-	if err != nil {
-		return nil, err
-	}
 	configurations, err := GetConfigurations()
 	if err != nil {
 		return nil, fmt.Errorf("failed to get configurations: %v", utils.ParseHarborErrorMsg(err))
 	}
 	robotNamePrefix := configurations.Payload.RobotNamePrefix.Value
 	queryName := ""
+	filter := ""
 	// Remove prefix from targetRobotName if it exists in the name
 	// Otherwise, use the targetRobotName as is
 	// If there is an error in the targetRobotName, it will be caught later when no robots are found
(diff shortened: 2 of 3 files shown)


pkg/api/robot_handler.go
```diff
@@ -61,6 +62,50 @@ func GetRobot(robotID int64) (*robot.GetRobotByIDOK, error) {
...
+		listResponse, err = ListRobot(ListFlags{})
+		if err != nil {
+			return nil, fmt.Errorf("failed to list robots: %v", utils.ParseHarborErrorMsg(err))
+		}
+	}
+
+	robotID := int64(-1)
+	for _, robotItem := range listResponse.Payload {
+		if robotItem.Name == targetRobotName {
+			robotID = robotItem.ID
```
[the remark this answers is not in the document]
Vad1mo: @qcserestipy ?

cmd/harbor/root/project/robot/delete.go
```diff
@@ -63,10 +66,20 @@ Examples:
 				err     error
 			)
 			if len(args) == 1 {
-				robotID, err = strconv.ParseInt(args[0], 10, 64)
+				if ProjectName == "" {
+					projectID := prompt.GetProjectIDFromUser()
+					project, err := api.GetProject(strconv.FormatInt(projectID, 10), true)
+					if err != nil {
+						log.Fatalf("failed to get project by name %s: %v", ProjectName, utils.ParseHarborErrorMsg(err))
```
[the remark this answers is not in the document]
Vad1mo: @qcserestipy ?

pkg/api/robot_handler.go
```diff
@@ -61,7 +62,63 @@ func GetRobot(robotID int64) (*robot.GetRobotByIDOK, error) {
...
+	} else {
+		listResponse, err = ListRobot(ListFlags{
+			Q: "name=" + queryName,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to list robots: %v", utils.ParseHarborErrorMsg(err))
+		}
+	}
+	if len(listResponse.Payload) > 1 {
+		return nil, fmt.Errorf("multiple robots found with name: %v", targetRobotName)
```
bupd, line 104: In harbor we usually select the first robot account form the list can you check if this works against same robot account name in both project and system level. 

This might be a edge case.

pkg/api/robot_handler.go
```diff
@@ -61,7 +62,63 @@ func GetRobot(robotID int64) (*robot.GetRobotByIDOK, error) {
...
+	} else if len(listResponse.Payload) == 0 {
+		return nil, fmt.Errorf("no robot found with name: %v, does it exist?", targetRobotName)
+	}
+	response, err := client.Robot.GetRobotByID(ctx, &robot.GetRobotByIDParams{RobotID: listResponse.Payload[0].ID})
```
bupd: I think we don't need to do the roundtrip again since in the list response already contains the robot details.

bupd (changes requested):
@qcserestipy Thanks for your contribution 

Added my suggestions for possible edge cases and improvements.

pkg/api/robot_handler.go
qcserestipy: Yes, you are absolutely right, I'll remove this.

pkg/api/robot_handler.go
qcserestipy: I checked that and it works, that we can separate them. In the project query the project id is required as well as the project name prefix so they are unique in both cases.

qcserestipy:
@bupd Thank you for your review! I incorporated the changes accordingly. In case you have time feel free to have another look.

Issue #556, opened 2025-10-07
Pipeline fails on vulnerability-check.


Vad1mo:
Our curreng GH action pipeline fails on [vulnerability-check](https://github.com/goharbor/harbor-cli/actions/runs/18254759794/job/51974519509#logs)

gkbishnoi07:
Thank you for the response. I wasn’t aware of that, now im start working on it!

bupd:
This is already solved

PR #565, opened 2025-10-28
fix(ci): resolve pipeline failure in vulnerability-check by handling build gracefully

gkbishnoi07:
The pipeline was failing during the vulnerability-check stage, not because of an actual vulnerability, but due to how Dagger executed the check.
Dagger’s containerized environment attempted to run govulncheck inside a controlled context that expected a pre-existing Go module build, but the step failed early when the Go binary (harbor) wasn’t built or exited with a non-zero code.

Screenshot 2025-10-29 030134

fix: #556

### DIFF
--- a/.github/workflows/default.yaml
+++ b/.github/workflows/default.yaml
@@ -78,15 +78,18 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Dagger Version
-        uses: sagikazarmark/dagger-version-action@v0.0.1
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.22.x'
 
       - name: Run Vulnerability Check
-        uses: dagger/dagger-for-github@v7
-        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/default.yaml
```diff
@@ -78,24 +78,26 @@ jobs:
...
-        with:
-          version: ${{ steps.dagger_version.outputs.version }}
-          verb: call
-          args: vulnerability-check-report export --path=vulnerability-check.report
+        run: |
+          go build -o ./tmp/harbor ./cmd/harbor || true
+
+          go install golang.org/x/vuln/cmd/govulncheck@v1.1.4
+
+          $(go env GOPATH)/bin/govulncheck -mode=binary ./tmp/harbor | tee vulnerability-check.report || true
```
bupd, line 92: We are using dagger and the workflow already works fine.

gkbishnoi07: okay i thought this issue still present

PR #607, opened 2026-01-20
fix : add url validation to all commands

### LINKED ISSUE #606: Add URL validation to all commands that accept URL input
## Summary
Several CLI commands accept URL inputs but lack proper validation, leading to confusing API errors when invalid URLs are provided.

## Problem
Commands like `instance create`, `registry create/update`, `scanner create/update`, `ldap ping`, and `webhook edit` accept URLs via flags but don't validate them before making API calls. Users see cryptic server errors instead of clear validation messages.

## Solution
Add `utils.ValidateURL()` validation to all commands that accept URL inputs:
- `instance create` - validate `--url` flag
- `registry create` - validate `--url` flag  
- `registry update` - validate `--url` flag
- `scanner create` - validate `--url` flag
- `scanner update` - validate `--url` flag
- `ldap ping` - validate `--ldap-url` flag
- `webhook edit` - validate `--endpoint-url` flag

Also update corresponding view files to use consistent URL validation in interactive forms.

Sypher845:
Fixes #606 
Adds URL validation to commands that accept URL inputs (`instance create`, `registry create/update`, `scanner create/update`, `webhook edit`).

Changes:
- Added `utils.ValidateURL()` checks before API calls
- Updated view files to use consistent URL validation in forms 

### ValidateURL function

What it checks:

* Valid URL format (scheme, host, etc.)
* Host is not empty
* Host is a valid IP or domain name
* Starts/ends with alphanumeric
* Hyphens allowed in middle
* Max 63 chars per label
* TLD at least 2 letters

````
func ValidateURL(rawURL string) error {
	var domainNameRegex = regexp.MustCompile(`^(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,}$`)

	parsedURL, err := url.ParseRequestURI(rawURL)
	if err != nil {
		return fmt.Errorf("invalid URL format: %v", err)
	}

	host := parsedURL.Hostname()
	if host == "" {
		return fmt.Errorf("URL must contain a valid host")
	}

	if net.ParseIP(host) != nil {
		return nil
	}

	if !domainNameRegex.MatchString(host) {
		return fmt.Errorf("invalid host: must be a valid IP address or domain name")
	}

	return nil
}
````

### DIFF
--- a/cmd/harbor/root/instance/create.go
+++ b/cmd/harbor/root/instance/create.go
@@ -47,9 +47,11 @@ You will need to provide the instance's name, vendor, endpoint, and optionally o
 			}
 
 			if opts.Name != "" && opts.Vendor != "" && opts.Endpoint != "" {
-				if err := utils.ValidateURL(opts.Endpoint); err != nil {
+				formattedEndpoint := utils.FormatUrl(opts.Endpoint)
+				if err := utils.ValidateURL(formattedEndpoint); err != nil {
 					return err
 				}
+				opts.Endpoint = formattedEndpoint
 				err = api.CreateInstance(opts)
 			} else {
 				err = createInstanceView(createView)
--- a/cmd/harbor/root/login.go
+++ b/cmd/harbor/root/login.go
@@ -130,7 +130,7 @@ func RunLogin(opts login.LoginView) error {
 	}
 	err := utils.ValidateURL(opts.Server)
 	if err != nil {
-		return err
+		return fmt.Errorf("invalid server URL: %w", err)
 	}
 	client := utils.GetClientByConfig(clientConfig)
 	ctx := context.Background()
--- a/cmd/harbor/root/registry/create.go
+++ b/cmd/harbor/root/registry/create.go
@@ -46,10 +46,12 @@ func CreateRegistryCommand() *cobra.Command {
 			}
 
 			if opts.Name != "" && opts.Type != "" && opts.URL != "" {
-				err = utils.ValidateURL(opts.URL)
+				formattedUrl := utils.FormatUrl(opts.URL)
+				err = utils.ValidateURL(formattedUrl)
 				if err != nil {
 					return err
 				}
+				opts.URL = formattedUrl
 				err = api.CreateRegistry(opts)
 			} else {
 				err = createRegistryView(createView)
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

**Comment by Sypher845** [comment]
@bupd can you please review this

**Comment by Sypher845** [comment]
@bupd i have merged the main branch in this , can you please review again .
Also do i have to make any other changes?

File: pkg/views/scanner/update/view.go [hunk 1]
```diff
@@ -105,7 +105,16 @@ func UpdateScannerView(scanner *models.ScannerRegistration) {
 			huh.NewInput().
 				Title("Scanner Adapter URL").
 				Value(&url).
-				Validate(huh.ValidateNotEmpty()),
+				Validate(func(str string) error {
+					if str == "" {
```
NucleoFusion: Could use `.TrimSpace` like in `pkg/views/registry/update/view.go`

pkg/views/webhook/edit/view.go
```diff
@@ -104,7 +104,15 @@ func WebhookEditView(editView *EditView) {
 
 			huh.NewInput().Title("Endpoint URL").
 				Value(&editView.EndpointURL).
-				Validate(utils.EmptyStringValidator("Endpoint URL")),
+				Validate(func(str string) error {
+					if str == "" {
```
NucleoFusion: Could use `.TrimSpace` like in `pkg/views/registry/update/view.go`

cmd/harbor/root/scanner/update.go
```diff
@@ -93,7 +94,11 @@ Only the fields passed through flags will be updated; other fields will retain t
 				updateView.AccessCredential = opts.AccessCredential
 			}
 			if flags.Changed("url") {
-				updateView.URL = strfmt.URI(opts.URL)
+				formattedUrl := utils.FormatUrl(opts.URL)
+				if err := utils.ValidateURL(formattedUrl); err != nil {
+					return fmt.Errorf("invalid URL: %v", err)
+				}
```
NucleoFusion, line 100: The function seems to be returning a formatted error, so do we need to do it twice?

Sypher845: yes , no need for it  
it should  return err directly

bupd (approved):
please fix the suggestions from copilot - otherwise lgtm

PR #629, opened 2026-01-22
fix: make comment field optional in user create command

### LINKED ISSUE #628: `user create` command requires ALL fields including the optional `comment` field
###  Make comment field optional in user create command
The `user create` command requires ALL fields including the optional `comment` field for non-interactive mode.

Sypher845:
Fixes #628 
The `user create` command previously required ALL fields including the optional `comment` field for non-interactive mode. Now users can create accounts with only the required fields

### DIFF
--- a/cmd/harbor/root/user/create.go
+++ b/cmd/harbor/root/user/create.go
@@ -40,7 +40,7 @@ func UserCreateCmd() *cobra.Command {
 				Username: opts.Username,
 			}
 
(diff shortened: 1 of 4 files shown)


NucleoFusion (changes requested):
Could also probably add an `(optional)` to the description string. (for `-h` help)
Everything else works as intended.

bupd (changes requested):
@Sypher845 merge conflicts

PR #634, opened 2026-01-23
fix(user): `harbor user elevate` to handle empty responses

### LINKED ISSUE #633: `harbor user elevate` executes even in absence of users
# Overview

The `harbor-cli` continue to execute, even there is no user accounts in the registry. 

## Steps to reproduce 

1. Delete all the user accounts in the registry
2. Execute `harbor user elevate`

## Preview

Image

vg006:
# Overview

Fixes #633

The `elevate` sub-command of the `user` command is updated to,
1. Handle empty responses(*payload with no user a/c details*) from registry.
2. Handle the errors and log them properly.

cc: @bupd @NucleoFusion 

## Preview
demo

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,7 +270,7 @@ Harbor <2.0.0 is not supported.
 * **User Group:** Join Harbor user email group: [harbor-users@lists.cncf.io](https://lists.cncf.io/g/harbor-users) to get update of Harbor's news, features, releases, or to provide suggestion and feedback.
 * **Developer Group:** Join Harbor developer group: [harbor-dev@lists.cncf.io](https://lists.cncf.io/g/harbor-dev) for discussion on Harbor development and contribution.
 * **Slack:** Join Harbor's community for discussion and ask questions: [Cloud Native Computing Foundation](https://slack.cncf.io/), channel: [#harbor](https://cloud-native.slack.com/messages/harbor/), [#harbor-dev](https://cloud-native.slack.com/messages/harbor-dev/) and [#harbor-cli](https://cloud-native.slack.com/archives/C078LCGU9K6).
-* **Community Calls:** Every Tuesday at 15:00 CET/CEST or 19:30 IST - [Join Meeting](https://zoom.us/j/99658352431)
+* **Community Calls:** Every Tuesday at 15:00 CET/CEST or 19:30 IST - [Join Meeting](https://zoom.us/j/99658352431). View [Meeting Notes](https://hackmd.io/@YvLTAInfRdO2rlxkLg9G4w/S1QeYi_uxx) for details.
 
 # License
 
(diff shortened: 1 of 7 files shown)


vg006:
@qcserestipy As you can see [here](https://github.com/goharbor/harbor-cli/blob/main/cmd/harbor/root/cmd.go#L74-L80), the logger is set with `io.Discard` as a output writer in non-verbose mode, neither used `os.StdOut` nor `os.StdErr` writers. 

So this silences the logrus. I agree that it is logged in the verbose mode, but normally a command should provide a proper closure upon the exit right. So I set it explicitly using the Cobra's writers within the command itself, instead of tweaking the global configuration.

cmd/harbor/root/user/delete.go
```diff
@@ -30,6 +30,7 @@ func UserDeleteCmd() *cobra.Command {
 		Args:  cobra.MinimumNArgs(0),
 		Run: func(cmd *cobra.Command, args []string) {
 			// If there are command line arguments, process them concurrently.
+			log.SetOutput(cmd.OutOrStderr())
```
bupd: not required. Please remove - we dont use stdErr here.

pkg/utils/utils.go
```diff
@@ -196,23 +196,21 @@ func Capitalize(s string) string {
...
+		log.Fatal(err)
+	}
+	ctx := context.Background()
+	response, err := client.User.ListUsers(ctx, &user.ListUsersParams{})
+	if err != nil {
+		log.Fatal(err)
+	}
+	userId, err := uview.UserList(response.Payload)
+	if err != nil {
+		log.Fatal(err)
```
bupd, line 221: please avoid using Fatal.

vg006: `log.Fatal` is not introduced in this change, it is presented all across the file.

cmd/harbor/root/user/delete.go
vg006: Removed it.

qcserestipy:
@vg006 @bupd I think changing the styling and color for this command only is not really necessary. Since we return simple text in every other command it could be confusing that it is different here. Just returning the error message as text should be the simplest. I guess `fmt.Errorf` and ` fmt.Printf` should be enough.

vg006:
I thought of improving the UX and I have just used the styles that are already defined in the project. Just added a utility function that returns the given text, wrapped by ANSI sequences.

Fine, then as per your suggestion, let me make this PR minimal. 
But I will propose this type of styled messages in my ongoing work #635, which will have a good impression on the UI and improves usability.

cmd/harbor/root/user/delete.go
bupd: please remove this - not really needed.

vg006: Done

PR #639, opened 2026-01-23
fix: project view --id flag not working

### LINKED ISSUE #638: harbor project view --id flag is defined but not implemented
Image

The `--id` parameter is declared but remains unimplemented, so searching by ID currently fails.

Sypher845:
Fixes #638 
t-rec

1. In `pkg/utils/error.go`, the function `ParseHarborErrorCode` can now parse error code of type `(status CODE)`, before it was only able to parse error of type `[METHOD /path][CODE]`.

2. Removed the `CheckProject` function in `pkg/api/project_handler.go` and replaced the logic where it is used with `GetProject` 

3. Before the --id flag for project view was not working , now it correctly passes `isID` to `GetProject` and works as expected.


pkg/api/project_handler.go
```diff
@@ -204,14 +204,25 @@ func LogsProject(projectName string) (*project.GetLogsOK, error) {
 	return response, nil
 }
 
-func CheckProject(projectName string) (bool, error) {
+func CheckProject(projectNameOrID string, useProjectID bool) (bool, error) {
```
qcserestipy: The current `CheckProject` implementation largely duplicates the logic of `GetProject`, but in a more complex way.
When `useProjectID` is true, `GetProject` is already called internally, so `CheckProject` does not add any additional value in this case.
For name-based checks, the function performs a `HeadProject` call that ultimately serves the same purpose as `GetProject`: determining whether the project exists.

To reduce duplication and simplify the code path, we can remove `CheckProject` entirely and replace its usages with a direct existence check based on `GetProject`.

cmd/harbor/root/project/member/create.go
```diff
@@ -43,7 +43,7 @@ func CreateMemberCommand() *cobra.Command {
 			var err error
 
 			if len(args) > 0 {
-				ok, checkErr := api.CheckProject(args[0]) // verifying project name
+				ok, checkErr := api.CheckProject(args[0], isID) // verifying project name
```
qcserestipy: I think it would be more simple if this check is performed with the ` api.GetProject` function.

cmd/harbor/root/project/member/delete.go
```diff
@@ -40,7 +40,7 @@ func DeleteMemberCommand() *cobra.Command {
 			var err error
 
 			if len(args) > 0 {
-				ok, checkErr := api.CheckProject(args[0]) // verifying project name
+				ok, checkErr := api.CheckProject(args[0], isID) // verifying project name
```
qcserestipy: I think it would be more simple if this check is performed with the ` api.GetProject` function.

cmd/harbor/root/project/logs.go
```diff
@@ -53,7 +53,7 @@ func LogsProjectCommmand() *cobra.Command {
 			}
 
 			log.Debugf("Checking if project '%s' exists...", projectName)
-			projectExists, err := api.CheckProject(projectName)
+			projectExists, err := api.CheckProject(projectName, false)
```
qcserestipy: I think it would be more simple if this check is performed with the ` api.GetProject` function.

cmd/harbor/root/project/view.go
```diff
@@ -49,14 +49,14 @@ func ViewCommand() *cobra.Command {
 			}
 
 			log.Debugf("Checking existence of project: %s", projectName)
-			projectExists, err := api.CheckProject(projectName)
+			projectExists, err := api.CheckProject(projectName, isID)
```
qcserestipy: I think it would be more simple if this check is performed with the ` api.GetProject` function.

Sypher845:
@qcserestipy  
I found one thing , in `pkg/utils/error.go` the `ParseHarborErroCode` can only check for error of format `[404]`.
But the  `project, err = api.GetProject(projectName, isID)` in this the `err` has a format of `(status 404)`, the current `ParseHarborErrorCode` cannot parse this format, so i added the format checking also.
please take a look

qcserestipy:
Nice, thank you! LGTM

pkg/utils/error.go
```diff
@@ -53,13 +54,23 @@ func ParseHarborErrorMsg(err error) string {
...
 	if len(parts) >= 2 {
 		codePart := strings.TrimSpace(parts[1])
 		if strings.HasPrefix(codePart, "[") && len(codePart) == 4 {
 			code := codePart[1:4]
 			return code
 		}
 	}
+
+	// Try format: (status CODE) - e.g., (status 404)
+	re := regexp.MustCompile(`\(status\s+(\d{3})\)`)
```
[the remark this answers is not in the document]
Sypher845: The function is not called frequently. Also many other files are compiling the regex inside the function.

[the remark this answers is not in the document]
Sypher845: SHould i implement this suggestion? @bupd @qcserestipy

PR #645, opened 2026-01-24
Fix/add error handling to build dev

### LINKED ISSUE #644: Incomplete error-handling in `build-dev` dagger function
I was testing out the new approach to build harbor-cli from source, which uses dagger commands, and they were not working out for me.

### Error faced
I picked up the dagger command from the README:

Image

- This got fixed quickly by adding the source flag in the command, after which I got the following error : 

Image

- As visible in the screenshot, The dagger command did not return any binary, and was neither giving any error as to why it was failing.

- This is because of the following portion in the code : 

Image

The build-dev command does not return any error log, when the init function is not working.
After making the relevant changes, I found out that my downstream had not cloned any of the tags, because of which the command:
```
git describe --tags --abbrev=0
```
was failing and hence everything was failing. On fixing this the build command was working successfully.

adityachopra29:
# Overview
- Fixes #644 
- Adds error handling to `build-dev` dagger function. 
- Adds missing `--source` flag in the readme for `build-dev` function's README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,7 +244,7 @@ Make sure you have the latest [Dagger](https://docs.dagger.io/) installed in you
 
 ```bash
 git clone https://github.com/goharbor/harbor-cli.git && cd harbor-cli
-dagger call build-dev --platform darwin/arm64 export --path=./harbor-cli
+dagger call build-dev --source=. --platform darwin/arm64 export --path=./harbor-cli
 ./harbor-dev --help
 ```
 


README.md
```diff
@@ -244,7 +244,7 @@ Make sure you have the latest [Dagger](https://docs.dagger.io/) installed in you
 
 ```bash
 git clone https://github.com/goharbor/harbor-cli.git && cd harbor-cli
-dagger call build-dev --platform darwin/arm64 export --path=./harbor-cli
+dagger call build-dev --source=. --platform darwin/arm64 export --path=./harbor-cli
```
NucleoFusion, line 247: The source flag is not required anymore.
It has a default value.

adityachopra29: Okay..closing this PR then

PR #585, opened 2025-12-21
Create dagger function to run CLI tests

adityachopra29:
## Summary
- Creates dagger function to run tests using a dagger service running harbor
- Fixes #249 

## Changes
- Creates `harbor-test` dagger function which runs tests after starting a service running harbor
- Current tests supported are only login_tests (since currently only call the demo server)
- Creates documenation for the new command

## Implementation details
- Uses the same config for the harbor service as used in harbor-satellite [see here](https://github.com/container-registry/harbor-satellite/tree/main/test/e2e)
- Current harbor service starts the following component services :
   - Core (since it contains the main API)
   - Postgres (to strore container and credentials etc)
   - Redis server (to store sessions details and caches)
- Modifies the tests to initially call the container endpoints instead of the demo server

## Testing
- Currently the postgres and redis images used in running the harbor service are not uploaded to harbor-registry . (They are the same ones used by harbor-satellite). I have raised issues regarding them in harbor-satellite repo as well see [here](https://github.com/container-registry/harbor-satellite/issues/211#issuecomment-3671552933) and [here](https://github.com/container-registry/harbor-satellite/issues/210#issuecomment-3671534203).
- Due to this complete testing was not done. However, individual unit tests did not break with the new changes
## Usage
```
dagger call harbor-test
```

PR #648, opened 2026-01-25
Feat/harbor setup podman

### LINKED ISSUE #249: Make tests not Dependent on Demo Server

adityachopra29:
# Overview 
- As discussed in https://github.com/goharbor/harbor-cli/pull/585 (comment), instead of using dagger to spin up an instance of harbor in the CI, we would run a podman instance on which tests could be run. This is the implementation for the same.
- Fixes #249 

# Changes made
- Creates a new workflow (harbor-e2e) which setups Harbor instance and integrated into default workflow
- Uses the repo https://github.com/bupd/harbor-podman, but is modified to be use for non SELinux such as Ubuntu (since github actions runner prefer ubuntu)
- Modifies the e2e tests to use the local harbor instance endpoints, and have `demo.goharbor.io` as fallback

# Testing
- All changes are tested and are working on the CI succesfully (check CI)


test/helper/helpers.go
```diff
@@ -14,17 +14,118 @@
...
+	}
+}
+
+// isHarborHealthy checks if a Harbor instance is responding to health checks.
+func isHarborHealthy(baseURL string) bool {
+	client := &http.Client{
+		Timeout: 5 * time.Second,
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true, // #nosec G402 - Allow self-signed certs for local test instance only
```
adityachopra29, line 133: Note : Since we are using a local harbor setup using a self-signed certificate, and this helper function is used for it only, it was necessary to add this directive here (to mute lint error).

test/helper/helpers.go
```diff
@@ -14,17 +14,118 @@
...
+}
+
+const (
+	// Environment variables for local Harbor instance (set by CI workflow)
+	EnvHarborURL      = "HARBOR_URL"
+	EnvHarborUsername = "HARBOR_USERNAME"
+	EnvHarborPassword = "HARBOR_PASSWORD"
+
+	// Fallback to demo.goharbor.io
+	FallbackURL      = "https://demo.goharbor.io"
```
bupd: feel free to remove this .

test/helper/helpers.go
```diff
@@ -14,17 +14,118 @@
...
+		// Local instance typically runs on a single address
+		return []string{cfg.URL}
+	}
+
+	// For demo.goharbor.io, test multiple URL formats
+	return []string{
+		"http://demo.goharbor.io:80",
+		"https://demo.goharbor.io:443",
+		"http://demo.goharbor.io",
+		"https://demo.goharbor.io",
```
bupd: please replace this with harbor.local

bupd (changes requested):
lgtm 

please update with the changes. Thanks

cmd/harbor/root/login_test.go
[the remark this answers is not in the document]
adityachopra29: I think this account only exists on demo.goharbor.io . In that case it does not make sense to port this test. So should I create this account on the local instance or keep this test on demo.goharbor.io?

.github/workflows/harbor-e2e.yaml
```diff
@@ -0,0 +1,176 @@
...
+      harbor_url: ${{ steps.set-outputs.outputs.harbor_url }}
+      harbor_username: ${{ steps.set-outputs.outputs.harbor_username }}
+      harbor_password: ${{ steps.set-outputs.outputs.harbor_password }}
+    steps:
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y \
+            podman podman-docker podman-compose \
+            git tar gzip wget openssl curl golang-go
```
qcserestipy: `golang-go` from apt is unnecessary, the `actions/setup-go` step later handles the Go installation from `go.mod`.

.github/workflows/harbor-e2e.yaml
```diff
@@ -0,0 +1,176 @@
...
+          # Install yq for reliable YAML manipulation
+          sudo wget -qO /usr/local/bin/yq https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64
```
qcserestipy: yq downloaded without version pin or checksum

This pulls whatever `latest` yq is at the time, with no integrity check. Pin to a specific version and verify the checksum to avoid supply-chain risk:

​```yaml
YQ_VERSION=<VERSION>
YQ_CHECKSUM=<CHECKSUM>
sudo wget -qO /usr/local/bin/yq "https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_linux_amd64"
echo "${YQ_CHECKSUM}  /usr/local/bin/yq" | sha256sum -c -
sudo chmod +x /usr/local/bin/yq
​```

.github/workflows/harbor-e2e.yaml
```diff
@@ -0,0 +1,176 @@
...
+  setup-harbor:
+    runs-on: ubuntu-latest
+    env:
+      HARBOR_HOSTNAME: ${{ inputs.harbor_hostname }}
+      HARBOR_ADMIN_PASSWORD: ${{ secrets.harbor_admin_password }}
+      TERM: xterm  # Avoid tput errors from install.sh color output
+    outputs:
+      harbor_url: ${{ steps.set-outputs.outputs.harbor_url }}
+      harbor_username: ${{ steps.set-outputs.outputs.harbor_username }}
+      harbor_password: ${{ steps.set-outputs.outputs.harbor_password }}
```
qcserestipy: `harbor_password` is set as a plain-text job output. GitHub Actions does not mask arbitrary outputs in logs (only registered secrets are masked). Since the tests run in the same job anyway, these outputs aren't consumed by any downstream job. Either:
- Remove the outputs entirely (they're unused), or
- If you plan to split into multiple jobs later, pass the password via a secret

.github/workflows/harbor-e2e.yaml
```diff
@@ -0,0 +1,176 @@
...
+    secrets:
+      harbor_admin_password:
+        required: false # If no password is given, a random base64 password is generated at runtime
+
+jobs:
+  setup-harbor:
+    runs-on: ubuntu-latest
+    env:
+      HARBOR_HOSTNAME: ${{ inputs.harbor_hostname }}
+      HARBOR_ADMIN_PASSWORD: ${{ secrets.harbor_admin_password }}
```
qcserestipy: I think it would be unnecessary to expose the secret as env var since it will be accessible as secret anywhere in the pipeline.

.github/workflows/harbor-e2e.yaml
```diff
@@ -0,0 +1,176 @@
...
+          fetch-depth: 0
+
+      - name: Set up Go
+        if: ${{ inputs.run_tests }}
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+          cache: true
+
+      - name: Run E2E Tests
```
qcserestipy, line 169: It could be great to add some step that collects some logs so it provides diagnostics if a test fails. It could looks something like this, please check out whether it makes sense

​```yaml
- name: Collect Harbor logs on failure
  if: failure()
  run: |
    sudo podman ps -a
    sudo podman logs harbor-core 2>&1 | tail -200
    sudo podman logs harbor-jobservice 2>&1 | tail -100
​```

.github/workflows/harbor-e2e.yaml
```diff
@@ -0,0 +1,176 @@
...
+          echo "harbor_url=https://${HOST}" >> $GITHUB_OUTPUT
+          echo "harbor_username=admin" >> $GITHUB_OUTPUT
+          echo "harbor_password=${ADMIN_PW}" >> $GITHUB_OUTPUT
+
+      # E2E tests run in the same job to access the Harbor instance
+      - name: Checkout harbor-cli repo
+        if: ${{ inputs.run_tests }}
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
qcserestipy, line 160: `fetch-depth: 0` fetches full history, but on forks the tags from upstream may not exist. The Dagger `init()` in `.dagger/main.go` runs `git describe --tags --abbrev=0` and will fail with `fatal: No names found` if no tags are present. This isn't an issue on the upstream repo, but will break for any contributor testing on their fork.

Consider either:
- Documenting this requirement, or
- Making the Dagger init resilient to missing tags

qcserestipy (changes requested):
Thank you very much for the contribution! Nice foundation for the E2E infrastructure. A few things need to be addressed before merging.

qcserestipy (commented):
I was also wondering, if we decide to go with a podman based approach like here, whether it could be a good idea migrating most part of the compose setup from @bupd's repo here into a dedicated subfolder. In this way the configuration of the test environment could be more closely tied to this code and would not live in some upstream repo. Most likely some of the config modifications that are happening in the pipeline could also be avoided then.

qcserestipy (commented):
Also tagging @NucleoFusion here

.github/workflows/harbor-e2e.yaml
adityachopra29: I have incorporated git describe now here.  If git describe finds no tags (e.g. on a fork), it silently falls back to v0.0.0-dev instead of exiting non-zero

.github/workflows/harbor-e2e.yaml
adityachopra29: Yup it makes sense. Have added a seperate step for the same.

adityachopra29:
Hey @qcserestipy @NucleoFusion . I have made all the changes you suggested...they made sense.
As for migrating the compose setup from @bupd's repository, my thought is that sinse that is a project which can/should be used in different applications as well, would it make sense to copy/clone it in each of those applications ( harbor-satellite, for example, might also need the setup). 
So keeping it as a seperate repository makes sense in that way. What are you thoughts?
If this will require more discussion, I suggest merging this PR first so that others who dont want to depend on demo.goharbor.io can start using it, and including that repo as part of another PR.

bupd:
invalid and has conflicts

Issue #591, opened 2026-01-11
[CI/Test] Incrementally improve unit test coverage


rkcoder101:
**Description**
Currently, the code coverage for the project stands at approximately **11%** (according to Codecov). To ensure the stability of the CLI, prevent regressions, and facilitate safer refactoring in the future, we should aim to increase this coverage.

**Proposal**
An incremental approach to improving the test suite is proposed. Rather than attempting a massive overhaul at once, the effort can be broken down into smaller, manageable chunks based on packages or functionality.

**Goals**
* Identify critical paths that are currently untested.
* Add unit tests for utility functions and command logic.

rkcoder101:
Hi @bupd @Vad1mo I’d like to start improving test coverage by writing tests for the docs generator code. I will share my progress on the way and would ask for help if I need it at some point. Thanks!

PR #653, opened 2026-01-27
test/Unit tests for user command

rkcoder101:
## Description
This PR introduces unit tests for the user command (`cmd/harbor/root/user`) and user view (`pkg/views/user`)

**Relates to #591**

## Test Coverage Checklist
- [x] user list 
- [x] user create
- [x] user delete
- [x] user elevate
- [x]  user password

## Note

**No** changes have been done to the core logic of the application, only a little refactoring for the purpose of writing some nice and clean test code :)

## List of refactoring changes done till now  (for writing unit tests)
1.  `cmd/harbor/root/user/list.go`: Separated `PrintUser()` and `GetUsers()` from  `UserListCmd()`
2.  `pkg/utils/helper.go`: Added `FPrintFormat()` to support `io.Writer` injection. The original `PrintFormat()` remains unchanged, ensuring full backward compatibility with existing code.
3.  `pkg/utils/utils.go`: Added `FPrintPayloadInJSONFormat()` and `FPrintPayloadInYAMLFormat()` to support `io.Writer`. The original `PrintPayloadInJSONFormat()` and `PrintPayloadInYAMLFormat()` remain unchanged.
4. `pkg/views/user/list/view.go`: Separated `MakeUserRows()` from  `ListUsers()`
5. `cmd/harbor/root/user/create.go`: Separated `CreateUser()` from `UserCreateCmd()`
6. `cmd/harbor/root/user/create.go`: Removed the redundant `createUserView()` as it was just a wrapper which called two functions
7. `cmd/harbor/root/user/delete.go`: Separated `DeleteUser()` from `UserDeleteCmd()`
8. `cmd/harbor/root/user/elevate.go`: Separated `ElevateUser()` from `ElevateUserCmd()`
9. `cmd/harbor/root/user/elevate.go`: Separated `ChangePassword()` from `UserPasswordChangeCmd()`

### DIFF
--- a/cmd/harbor/root/user/create.go
+++ b/cmd/harbor/root/user/create.go
@@ -47,7 +47,9 @@ func CreateUser(opts *create.CreateView, userCreator UserCreator) {
 
 	if err != nil {
 		if isUnauthorizedError(err) {
-			log.Error("Permission denied: Admin privileges are required to execute this command.")
+			log.WithFields(log.Fields{
+				"action": "user create",
+			}).Error("Permission denied: The current account does not have the required permissions to create users.")
 		} else {
 			log.Errorf("failed to create user: %v", err)
 		}
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -18,7 +18,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"os"
 	"regexp"
 	"strconv"
@@ -35,7 +34,8 @@ import (
 )
 
 // Returns Harbor v2 client for given clientConfig
-func FPrintPayloadInJSONFormat(w io.Writer, payload any) {
+
+func PrintPayloadInJSONFormat(payload any) {
 	if payload == nil {
 		return
 	}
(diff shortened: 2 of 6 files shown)


pkg/utils/utils.go
```diff
@@ -45,10 +45,10 @@ func PrintPayloadInJSONFormat(payload any) {
 		panic(err)
 	}
 
-	fmt.Println(string(jsonStr))
+	fmt.Fprint(w, string(jsonStr))
```
bupd:
```suggestion
	fmt.Println(string(jsonStr))
```

revert this back - our usecase doesn't fit for Fprint

cmd/harbor/root/user/create.go
```diff
@@ -23,6 +23,36 @@ import (
...
+
+	if opts.Email == "" || opts.Realname == "" || opts.Password == "" || opts.Username == "" {
+		userCreator.FillUser(opts)
+	}
+
+	err = userCreator.UserCreate(*opts)
+
+	if err != nil {
+		if isUnauthorizedError(err) {
+			log.Error("Permission denied: Admin privileges are required to execute this command.")
```
bupd: robot accounts can also create users - this is not limited to admins.

pkg/utils/utils.go
rkcoder101: Oops I realised Fprint does not give a new line character at the end (like Println). Can I use Fprintln which enables both writer injection (and helps in testing) as well as gives `\n` at the end?
Actually I saw that PrintFormat() function is being used in many commands [(example)](https://github.com/goharbor/harbor-cli/blob/58a7157fa8f48a67574d07ee5bd21c25530fbd10/cmd/harbor/root/project/list.go#L95), and it would enable easy testing if we had a 'F' version of the function, through which we can inject a writer/buffer to capture and test the output.

cmd/harbor/root/user/create.go
rkcoder101: Ok I will correct this :+1:

PR #654, opened 2026-01-27
docs(readme): fix docker run config mount prerequisite

Jaisheesh-2006:
## What this PR does / why we need it

When running Harbor CLI via Docker, the README mounts a host config file:

`-v $HOME/.config/harbor-cli/config.yaml:/root/.config/harbor-cli/config.yaml`

If `$HOME/.config/harbor-cli/config.yaml` does not exist yet, Docker may create it as a **directory** on the host, which then causes a "file vs directory" bind-mount error on subsequent runs.

## What this PR does

Adds a small prerequisite step to the README "Run using Container" section to create the config directory and an empty `config.yaml` file before running `docker run`:

- `mkdir -p $HOME/.config/harbor-cli`
- `touch $HOME/.config/harbor-cli/config.yaml`

This ensures the bind mount always targets a file and avoids the directory-creation pitfall.

## Which issue(s) this PR fixes

N/A

## Special notes for your reviewer

Docs-only change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,3 @@
-
 ![Harbor-CLI Logo_256px](https://github.com/user-attachments/assets/fa18e8f0-a2e4-4462-ab2d-446a88f9edb3)
 
 **Harbor CLI — a command-line interface for interacting with your Harbor container registry. A streamlined, user-friendly alternative to the WebUI, as your daily driver or for scripting and automation.**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,16 +59,26 @@ The project's first goal is to reach WebUI parity.
...
 Running Harbor CLI as a container is simple. Use the following command to get started:
 
+Prerequisite: ensure the config file exists on the host before running Docker:
+
+```bash
+mkdir -p $HOME/.config/harbor-cli
+touch $HOME/.config/harbor-cli/config.yaml
+```
+
+This prevents the "directory vs file" bind-mount error where Docker creates `config.yaml` as a directory if the file doesn't exist yet.
```
bupd, line 73: this is not needed. Since this more depends how docker works and mounts the volume

PR #656, opened 2026-01-27
fix: add missing error returns in command handlers

AnvayKharb:
fixed #455 
This pull request fixes a recurring issue in the CLI where errors were being logged but execution continued as if nothing went wrong. In several commands, log.Errorf() was used without returning the error, which caused the CLI to proceed with invalid or empty values and fail silently.

To address this, the PR replaces error logging with explicit error returns using fmt.Errorf, ensuring that failures immediately stop execution and are correctly surfaced to the user. To support this flow, affected commands were updated to use Cobra’s RunE instead of Run, which is the recommended pattern when commands can fail.

As part of this cleanup, the unused logrus imports were removed from the affected files, since error reporting is now handled through returned errors rather than logging.

Overall, this change makes CLI behavior more predictable, prevents silent failures, and aligns error handling across commands with standard Cobra and Go best practices.

### DIFF
--- a/cmd/harbor/root/artifact/delete.go
+++ b/cmd/harbor/root/artifact/delete.go
@@ -19,7 +19,6 @@ import (
 	"github.com/goharbor/harbor-cli/pkg/api"
 	"github.com/goharbor/harbor-cli/pkg/prompt"
 	"github.com/goharbor/harbor-cli/pkg/utils"
-	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


NucleoFusion:
A full overhaul of all commands that use `Run + log.Fatal` is being done in #679

bupd:
invalid, vague and has conflicts

Issue #655, opened 2026-01-27
CLI continues execution after log.Errorf() instead of returning error


AnvayKharb:
In multiple Harbor CLI commands, when an error occurs it is logged using log.Errorf() but the function does not return the error. Execution continues after the error, causing variables like projectName, repoName, or reference to be empty or invalid.

Because of this, the command does not fail immediately and instead proceeds with incorrect state, leading to silent failures and confusing downstream behavior. The CLI appears to run successfully even though a critical error has already occurred.

PR #659, opened 2026-01-27
655 issue handling

DhruvTotala:
This PR fixes issue #655, where some CLI commands continued executing after an error was logged using log.Errorf(). Because the error was not returned, the command would proceed with empty or invalid values, leading to silent and confusing failures.

What has changed:-

1. Updated affected commands to use RunE instead of Run so errors can be properly returned
2. Replaced log.Errorf() with fmt.Errorf() and returned the error to stop execution
3. Removed os.Exit() calls and relied on error propagation instead
4. Ensured commands fail immediately when an error occurs, preventing invalid state

These changes make the CLI behavior more predictable and align error handling with standard Cobra CLI practices.


bupd:
invalid and has conflicts

PR #660, opened 2026-01-28
fix(member): handle error in DeleteAllMember to prevent nil pointer p…

goyalpalak18:
## Description
I ran into a crash when using `harbor project member delete --all` while my Harbor credentials were expired. 

Instead of telling me the request failed, the CLI panicked with a nil pointer dereference. It turns out the `DeleteAllMember` function was ignoring errors from the API and trying to read the empty response anyway.

This PR fixes that by adding proper error checking. Now, if the API call fails (network timeout, 401, etc.), the CLI will print a readable error message and exit cleanly instead of crashing with a stack trace.

## Verification
I reproduced this by simulating a network error during the delete command.
- **Before:** CLI panicked with `runtime error: invalid memory address`.
- **After:** CLI printed `failed to list members: connection refused` and exited with code 1.

## Checklist
- [x] Panic is fixed.
- [x] Error messages are readable.
- [x] Verified happy path still works.

### DIFF
--- a/pkg/api/member_handler.go
+++ b/pkg/api/member_handler.go
@@ -15,6 +15,7 @@
 package api
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 
@@ -109,7 +110,7 @@ func DeleteAllMember(projectName string, xIsResourceName bool) error {
 	length := len(response.Payload)
 	if length < 1 {
 		log.Info("No members found in project")
-		return nil
+		return fmt.Errorf("no members found in project %s", projectName)
 	}
 
 	errChan := make(chan error, length)
@@ -140,7 +141,7 @@ func DeleteAllMember(projectName string, xIsResourceName bool) error {
 	}
 
 	if len(errs) > 0 {
-		return fmt.Errorf("failed to delete %d member(s): %v", len(errs), errs[0])
+		return fmt.Errorf("failed to delete %d member(s): %w", len(errs), errors.Join(errs...))
 	}
 
 	return nil
--- a/pkg/api/member_handler_test.go
+++ b/pkg/api/member_handler_test.go
@@ -53,7 +53,8 @@ func TestDeleteAllMember_NoMembers(t *testing.T) {
 
 	err := DeleteAllMember("test-project", true)
 
-	assert.NoError(t, err)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "no members found in project")
 }
 
 func TestDeleteAllMember_Success(t *testing.T) {


pkg/api/member_handler.go
```diff
@@ -92,40 +98,52 @@ func CreateMember(opts create.CreateView) error {
...
+	response, err := listMembersFunc(projectName, "", true)
+	if err != nil {
+		return fmt.Errorf("failed to list members: %w", err)
+	}
+
+	length := len(response.Payload)
 	if length < 1 {
 		log.Info("No members found in project")
-		return
+		return nil
```
NucleoFusion: This should return an error, that no members found alongside log

pkg/api/member_handler.go
```diff
@@ -92,40 +98,52 @@ func CreateMember(opts create.CreateView) error {
...
 	for err := range errChan {
 		if err != nil {
-			log.Errorln("Error:", err)
+			errs = append(errs, err)
 		}
 	}
+
+	if len(errs) > 0 {
+		return fmt.Errorf("failed to delete %d member(s): %v", len(errs), errs[0])
+	}
```
NucleoFusion, line 145: This should `errors.Join()` and not only the first error

pkg/api/member_handler_test.go
```diff
@@ -0,0 +1,113 @@
...
+	// Mock ListMembers to return empty list
+	listMembersFunc = func(projectNameOrID, memberName string, isName bool) (*member.ListProjectMembersOK, error) {
+		return &member.ListProjectMembersOK{
+			Payload: []*models.ProjectMemberEntity{},
+		}, nil
+	}
+
+	err := DeleteAllMember("test-project", true)
+
+	assert.NoError(t, err)
```
NucleoFusion: this assert would after changes give no members found error or something

pkg/api/member_handler.go
goyalpalak18: fixed, it returns an error now instead of nil

pkg/api/member_handler.go
goyalpalak18: good call. swapped it to use `errors.Join(errs...)`

pkg/api/member_handler_test.go
goyalpalak18: yep, updated the test to expect the error

NucleoFusion: @goyalpalak18 you can resolve review comments btw. That is easier to navigate for you and us as well.
Thanks for the contribution

PR #666, opened 2026-02-03
test: make harbor login tests deterministic using httptest server

### LINKED ISSUE #665: Stabilize `harbor login` tests by removing live network dependency
While exploring the test suite for the Harbor CLI, I noticed that the harbor login tests appear to rely on live network requests to https://demo.goharbor.io.

Specifically :
- Tests under `cmd/harbor/root/login_test.go` exercise logic in `cmd/harbor/root/login.go`
- Some paths perform real API calls against the demo Harbor instance

This can make the tests :
- flaky due to network or remote service availability
- slower than necessary
- harder to run deterministically in CI or local environments

**Proposed improvement**
Replace the live network dependency in `login` tests with a mocked HTTP client or a local `httptest.Server` that simulates the required Harbor API responses. This would keep the existing behavior intact while making the tests fast and deterministic.

Possible approaches (open to guidance) :
- Inject an HTTP client or base URL into the login flow for testing
- Use `httptest.Server` to mock the `/api/v2.0/users/current` (or relevant) endpoints
- Limit the scope to test-only changes without affecting production behavior

Why this matter's ?
- Aligns with best practices for CLI and API testing
- Improves CI stability and developer experience
- Makes test failures easier to reason about

rshdhere:
Closes #665

Tests verified locally using `dagger call test --source=.`

### DIFF
--- a/cmd/harbor/root/login_test.go
+++ b/cmd/harbor/root/login_test.go
@@ -14,6 +14,9 @@
 package root_test
 
 import (
+	"io"
+	"net/http"
+	"net/http/httptest"
 	"testing"
 
 	"github.com/goharbor/harbor-cli/cmd/harbor/root"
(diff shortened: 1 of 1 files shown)


rshdhere:
> From what I know about testing. Creating an `httptest` mockserver would be required to proofcheck how our command handles vague http responses. As far as relieving dependency on demo.goharbor.io goes, as stated in the original issue, **we want to create a local (github actions based) mock instance of harbor. And check through that.** and, If we want to make `httptest` based testing, it would be for handling vague http response codes / errors, such as network faults, internal server errors etc @qcserestipy what do you think?

@NucleoFusion That's a fair point about needing real Harbor validation. 

I see this PR as solving the immediate flakiness issue with unit tests, while what you're describing sounds like proper integration testing - which I agree we need.

Would it make sense to:
1. Merge these httptest mocks for fast, deterministic unit tests
2. Add GitHub Actions-based Harbor instance for integration tests (happy to help with this)

Or would you prefer I work on the GH Actions approach first instead?

NucleoFusion:
> I see this PR as solving the immediate flakiness issue with unit tests

We dont really need an immediate response, this is not a critical issue right now

NucleoFusion:
@qcserestipy absolutely, I believe we really need the instance before we do the official v1 release. And its a lot of reading (at least for me) on how Harbor Satellite / Podman does this.

Issue #667, opened 2026-02-03
Chore: Move types to relevant packages


bupd:
https://github.com/goharbor/harbor-cli/blob/main/pkg/api/types.go - We are planning to remove this file and move the types in this file to the relevant packages such as in the views package.

PR #669, opened 2026-02-03
chore(types): moving types to the relevent packages and deleting types.go

rshdhere:
solves #667 

tested using dagger locally!


rshdhere:
@bupd open for the review #667

pkg/views/context/types.go
```diff
@@ -0,0 +1,23 @@
...
+
+// Package contextview holds shared types for context-related views.
+package contextview
+
+// ContextListView represents a context entry used in list/switch views.
+type ContextListView struct {
+	Name     string
+	Username string
+	Server   string
+}
```
NucleoFusion, line 23: These can follow the existing structure and be included in the `view.go` itself.

pkg/views/registry/create/types.go
```diff
@@ -0,0 +1,31 @@
...
+
+// CreateView is the interactive view model to create a registry.
+type CreateView struct {
+	Name        string
+	Type        string
+	Description string
+	URL         string
+	Credential  RegistryCredential
+	Insecure    bool
+}
```
NucleoFusion, line 31: These can follow the existing structure and be included in the `view.go` itself.

PR #671, opened 2026-02-04
Validate refresh-interval to be greater than 0

### LINKED ISSUE #668: Add Validation for audit-logs refresh-interval to be greater than 0
## Description

https://github.com/goharbor/harbor-cli/blob/745b6ac3724459d8ceeab751397b441072ce44e0/cmd/harbor/root/logs.go#L120

Sbragul26:
### What this PR does,
- Adds validation to ensure `--refresh-interval` is greater than 0
- Prevents invalid values (like `0` or negative durations) that could cause tight loops

### Related issue
- Fixes #668

### DIFF
--- a/cmd/harbor/root/logs.go
+++ b/cmd/harbor/root/logs.go
@@ -60,6 +60,10 @@ harbor-cli logs --output-format json`,
 					if err != nil {
 						log.Fatalf("invalid refresh interval: %v", err)
 					}
+
+					if interval <= 0 {
+						log.Fatalf("refresh-interval must be greater than 0")
+					}
 				}
 				followLogs(opts, interval)
 			} else {


rshdhere:
Nice, this addresses the issue cleanly 👍 validating --refresh-interval upfront avoids accidental tight loops and makes the behavior safer.

Small UX thought (non-blocking): since Harbor CLI is Cobra-based, would it be worth returning this as a flag validation error (e.g. via PreRunE / returning an error) instead of log.Fatalf, so the CLI exits with a consistent usage/error message like other commands? what are your thoughts @NucleoFusion 

Not required for this PR — overall this looks straightforward and well-scoped.

cmd/harbor/root/logs.go
```diff
@@ -60,6 +60,10 @@ harbor-cli logs --output-format json`,
 					if err != nil {
 						log.Fatalf("invalid refresh interval: %v", err)
 					}
+
+					if interval <= 0 {
+						log.Fatalf("refresh-interval must be greater than 0")
+					}
```
NucleoFusion, line 66: Interval should be more than 0.5sec or something. 0 would make it block cpu resources.

PR #679, opened 2026-02-06
replace commands that use Run to use RunE

### LINKED ISSUE #675: refactor: Replace commands that use `Run` to use the `RunE` from cobra
Several older commands rely on a combination of `Run` + `log` to handle errors. In the newer commands, we have turned towards using the `RunE` field for cobra commands.

We need to go through and change all the `Run` instances to `RunE`

bhavyaKhatri2703:
fixes #675 
replaced older commands from Run to RunE with errors being returned properly instead of jusing log.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,11 @@
 version: 2
 updates:
-  - package-ecosystem: gomod
+  - package-ecosystem: "gomod"
     directory: "/"
     schedule:
-      interval: weekly
+      interval: "weekly"
+
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


rshdhere:
@bhavyaKhatri2703 I see your commit messages. it's looking good, but follow `contributing.md` which has proper format for commit messages

ig dont need for this PR but do consider it in the upcoming PRs

bhavyaKhatri2703:
> Please fix the ci - test-code pipeline is failing
> 
> otherwise LGTM
> 
> @bhavyaKhatri2703 Thanks for the contribution

fixed now.

PR #681, opened 2026-02-06
fix(prompt): remove GetTagNameFromUser() that caused deadlock

### LINKED ISSUE #680: bug: GetTagNameFromUser() causes `deadlock` due to empty goroutine
The function `GetTagNameFromUser()` in `pkg/prompt/prompt.go` maybe contains a critical bug that causes the program to deadlock (hang forever) when called.

The current implementation creates an `unbuffered channel` but the goroutine is completely empty, meaning no value is ever sent to the channel. This causes the function to block indefinitely when it attempts to receive from the channel.

```
func GetTagNameFromUser() string {
    repoName := make(chan string)
    go func() {
        // Empty - nothing sends to the channel
    }()
    return <-repoName  // Blocks forever - DEADLOCK
}
```

**Expected Behavior**
The function should prompt the user to select a tag name, similar to other prompt functions in the same file.

**Similar Working Pattern**
For reference, here's a correctly implemented function in the same file:

```
func GetScannerIdFromUser() string {
    scannerId := make(chan string)
    go func() {
        response, _ := api.ListScanners()
        sview.ScannerList(response.Payload, scannerId)  // Sends to channel
    }()
    return <-scannerId
}
```

**Suggested Fix**
Implement the function body to actually fetch tags and present them to the user:

```
func GetTagNameFromUser() string {
    tagName := make(chan string)
    go func() {
        // TODO: Implement tag selection logic
        // This needs additional parameters (project, repo, reference) 
        // to fetch tags from the API
    }()
    return <-tagName
}
```
**Note**:  This function may need to be redesigned to accept parameters (project name, repository name, reference) or be removed if it's redundant with the existing GetTagFromUser() function which already handles tag selection properly.

rshdhere:
solves #680 
The function had an empty goroutine that never sent to the unbuffered channel, causing any caller to block indefinitely. Since this function was unused and GetTagFromUser() already provides the correct implementation for tag selection, removing it entirely is the cleanest fix. Fixes #680


rshdhere:
PTAL @NucleoFusion

pkg/prompt/prompt.go
```diff
@@ -192,14 +192,9 @@ func GetTagFromUser(repoName, projectName, reference string) string {
...
 
-func GetTagNameFromUser() string {
-	repoName := make(chan string)
-
-	go func() {
-	}()
-
-	return <-repoName
-}
+// GetTagNameFromUser was removed - it contained a deadlock bug (empty goroutine
```
qcserestipy: I think we can remove it without leaving a comment. So far no one used it and will not miss it, I guess

rshdhere: sorry for the delay was busy with school, on it now!

PR #682, opened 2026-02-06
feat(user): add fuzzy, match, range flags for commands

### LINKED ISSUE #674: refactor: Refactor old commands that support the `q` query paramater to now utilize the new format.
Most commands use the `Q` for querying a list of results. There format is very non-user friendly. 

```
Query string to query resources. Supported query patterns are "exact match(k=v)", "fuzzy match(k=~v)", "range(k=[min~max])", "list with union releationship(k={v1 v2 v3})" and "list with intersetion relationship(k=(v1 v2 v3))". The value of range and list can be string(enclosed by " or '), integer or time(in format "2020-04-09 02:36:00"). All of these query patterns should be put in the query string "q=xxx" and splitted by ",". e.g. q=k1=v1,k2=~v2,k3=[min~max]
```

This is the official description.
Ex: `name=~some,id=222`
And in a command this would look like:-

```
harbor-cli user list --query name=~some,id=222
```
This is un-intuitive.

Instead for a more UX friendly approach we have a new solution in some of the commands. Which is,

```
harbor-cli user list --fuzzy name=some --match id=222 ---range <something>
```

And this is supported by,

```
func BuildQueryParam(fuzzy, match, ranges []string, validKeys []string) (string, error) {
```

In the `pkg/utils/query.go`.

With flag format like this,

```
	flags.StringSliceVar(&fuzzy, "fuzzy", nil, "Fuzzy match filter (key=value)")
	flags.StringSliceVar(&match, "match", nil, "exact match filter (key=value)")
	flags.StringSliceVar(&ranges, "range", nil, "range filter (key=min~max)")
```

Although this is implemented for a few of the new commands, the old ones still rely on the outdated format.
We need to update them to be similar to the new ones.

Mujib-Ahasan:
## Description 
Only `harbor-cli project list` and `harbor-cli label list` support fuzzy, match, and range flags. All the other commands does not follow this way to query resource. 
This PR has been raised to make fuzzy, match, range flag available to other required commands(eg: user, artifact, instance, registry etc). 

Resolves #674


NucleoFusion:
Have you done all those? In the checklist that have CLI present?
If so I will have a look ASAP

NucleoFusion:
Sorry for the delay.
I can see that even though some of the CLI Commands dont have `--query` builtin, but their API in Harbor does support it.
For example,

`tags/immutable/list.go, artifact/tags.go`

image

image

The list I have given has all the endpoints that support `q` param as per the API. not the CLI.
Please check and add to even those that done have the `--query` currently.
@Mujib-Ahasan

cmd/harbor/root/project/robot/list.go
```diff
@@ -112,13 +115,7 @@ Examples:
 	flags.Int64VarP(&opts.PageSize, "page-size", "", 10, "Size of per page")
 	flags.Int64VarP(&opts.ProjectID, "project-id", "", 0, "Project ID")
 	flags.StringVarP(&opts.Q, "query", "q", "", "Query string to query resources")
```
NucleoFusion, line 148: No query conversion here

NucleoFusion (changes requested):

NucleoFusion (approved):
lgtm

Mujib-Ahasan:
Hey @bupd, PTAL, whenever you get time!

bupd (changes requested):
@Mujib-Ahasan  Thanks for your contribution, Please fix CI and also rebase your commits and resolve conflicts.

PR #683, opened 2026-02-06
feat(user): add confirm password field to user create command

### LINKED ISSUE #673: enhancement: Refine the `user create` command flags and more
Currently, the user create command looks like this.

Image

The issues I recognize are:-

1. There is no 'Confirm Password' field. We should add a `Confirm Password` field.
2. The name of the comment field should add a `optional` indicator.
3. I do not know the plausibility of this but a toggle for viewing / not viewing password could be added.

Sypher845:
image

Fixes #673 
 1. Confirm password field added to the user create command 
 2. the comment field optional indicator is handled in #629 
 3. For the toggle,
views use `charmbracelet/huh` library. The `EchoMode` is set at form creation time and cannot be dynamically toggled during input.

### DIFF
--- a/pkg/views/user/create/view.go
+++ b/pkg/views/user/create/view.go
@@ -23,11 +23,12 @@ import (
 )
 
 type CreateView struct {
-	Username string
-	Email    string
-	Realname string
-	Comment  string
-	Password string
+	Username        string
+	Email           string
+	Realname        string
+	Comment         string
+	Password        string
+	ConfirmPassword string
 }
 
 func CreateUserView(createView *CreateView) {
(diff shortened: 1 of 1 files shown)


NucleoFusion:
for 2. I meant in the createView. Like the form you see in the image you have attached.
Adding an `(optional)` in here.
The other PR handles for the command commment iirc

Sypher845:
It is not possible,
As in the current use of huh library, the Ecomode is set at form creation, so it cannot be changed during input

Sypher845:
We are creating form , The Echomode for password which is set `EchoMode(huh.EchoModePassword)` is like keep it hidden ( like *****). After the `form.Run()` is called, the forms state become static, we cannot change this during the run or when we input :)

NucleoFusion:
Okay, I see
Its fine then

doc/cli-docs/harbor-user-create.md
```diff
@@ -15,12 +15,13 @@ harbor user create [flags]
...
 
 ```sh
-      --comment string    Comment
-      --email string      Email
-  -h, --help              help for create
-      --password string   Password
-      --realname string   Realname
-      --username string   Username
+      --comment string            Comment
+      --confirm-password string   Confirm Password
```
qcserestipy: I am not sure whether this should be introduced as an additional flag for the CLI command. I agree that having a password confirmation in interactive mode is useful, since the password is obfuscated there. However, for the CLI command, the `--password <password>` argument is visible in clear text in the terminal when creating a user, so a confirmation flag does not seem necessary in that case.

@NucleoFusion What do you think?

qcserestipy (changes requested):
Thank you for the contribution. Before approving this, we should consider whether a confirmation flag is actually needed for the CLI command.

Sypher845:
Changed the commit name.
I am not able to edit the PR name, as the edit button is not showing for me

PR #688, opened 2026-02-07
tests: Added unit tests for user create command

rkcoder101:
## Description
This PR introduces unit tests for the user create command (`cmd/harbor/root/user/create.go`)

**Relates to #591**

**Is a sub-part of #653**

## Note

**No** changes have been done to the core logic of the application, only a little refactoring for the purpose of writing some nice and clean test code :)

### DIFF
--- a/cmd/harbor/root/user/create.go
+++ b/cmd/harbor/root/user/create.go
@@ -23,6 +23,38 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type UserCreator interface {
+	FillUser(opts *create.CreateView)
+	UserCreate(opts create.CreateView) error
+}
+type DefaultUserCreator struct{}
+
+func (d *DefaultUserCreator) FillUser(opts *create.CreateView) {
+	create.CreateUserView(opts)
+}
+
+func (d *DefaultUserCreator) UserCreate(opts create.CreateView) error {
+	return api.CreateUser(opts)
+}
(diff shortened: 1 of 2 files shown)


qcserestipy (changes requested):
Thanks for the contribution! One concern with the current approach:

var fillUser = create.CreateUserView was added purely to allow tests to swap it out. This monkey-patching pattern modifies production code just for testability, which isn't necessary here.

CreateUser already accepts createUserAPI as a function parameter — the same pattern can be applied to fillView. Remove the package-level var and add fillView as a third parameter instead:

```go
func CreateUser(opts *create.CreateView, fillView func(*create.CreateView), createUserAPI func(opts create.CreateView) error) {
    if opts.Email == "" || opts.Realname == "" || opts.Password == "" || opts.Username == "" {
        fillView(opts)
    }
    // ...rest unchanged
}
```

Then in UserCreateCmd, pass the real implementation:
```go
CreateUser(&opts, create.CreateUserView, api.CreateUser)
```
And in the test, pass the mock directly — no need to touch the package-level var at all:
```go
CreateUser(opt, m.fillUser, m.userCreate)
```
This keeps create.go clean, removes the package-level var entirely, and is consistent with how createUserAPI injection is already done. All test helpers stay where they belong in create_test.go.

bupd:
stale and has conflicts

PR #690, opened 2026-02-07
refactor(robot): dynamically fetch and sort permissions with improved error handling

qcserestipy:
## Overview

This PR does not change the functionality, but refactors the `NewRobotPermissionsGrid` in the robot permission selection. Before it was relying on hard coded actions and permissions for the table creation. This information is now retrieved directly from the harbor api as the permission spec might change in the future and hard coded parts break. Additionally, the go routine passes a customn struct that allows for more cleaner error handling than just `os.Exit` on failure

## Changes
- Fetch available permissions and actions from API instead of using hardcoded values
- Sort resources and actions deterministically for reliable UI presentation
- Introduce PermissionSelectResult struct for proper error propagation in channel
- Support both 'project' and 'system' permission kinds
- Remove os.Exit() calls in favor of channel-based error handling

### DIFF
--- a/pkg/prompt/prompt.go
+++ b/pkg/prompt/prompt.go
@@ -319,7 +319,14 @@ func GetActiveContextFromUser() (string, error) {
 func GetRobotPermissionsFromUser(kind string) ([]models.Permission, error) {
 	permissions := make(chan robotView.PermissionSelectResult)
 	go func() {
-		response, _ := api.GetPermissions()
+		response, err := api.GetPermissions()
+		if err != nil {
+			permissions <- robotView.PermissionSelectResult{
+				Permissions: nil,
+				Err:         err,
+			}
+			return
+		}
 		robotView.ListPermissions(response.Payload, kind, permissions)
 	}()
 	result := <-permissions


pkg/prompt/prompt.go
```diff
@@ -316,13 +316,14 @@ func GetActiveContextFromUser() (string, error) {
 	return res, nil
 }
 
-func GetRobotPermissionsFromUser(kind string) []models.Permission {
-	permissions := make(chan []models.Permission)
+func GetRobotPermissionsFromUser(kind string) ([]models.Permission, error) {
+	permissions := make(chan robotView.PermissionSelectResult)
 	go func() {
 		response, _ := api.GetPermissions()
 		robotView.ListPermissions(response.Payload, kind, permissions)
```
NucleoFusion: Could we integrate the `api.GetPermissions` error to be returned here as well?

NucleoFusion (changes requested):
Looks pretty good!
We could also handle the `api.GetPermissions()` error here in my opinion

pkg/prompt/prompt.go
qcserestipy: Good hint! I added error handling for this one. In a next PR I will deal with all the unhandled errors in Robot account stuff in `prompt.go`

NucleoFusion (approved):
lgtm!

bupd (approved):
/lgtm

PR #708, opened 2026-02-16
refactor(robot): introduce listselect model for project prompts

qcserestipy:
## refactor(robot): introduce reusable listselect model for project prompts

Replaces the inline project selection logic in robot create/update with a reusable multi-select list model based on Bubble Tea.

### Changes
- Add `listselect` model with toggle (`space`) and confirm (`enter`) support
- Add `GetProjectNamesFromUser` and `GetProjectIDsFromUser` to `prompt` package
- Refactor system robot create/update to use new model

robot_create


qcserestipy (commented):
This PR is part of further refactoring the robot accounts code. The main point of this PR is to introduce the multiselect view which allows to select several items from a list. In the current implementation of the listselect model we can only select one single item from a list. This is to get rid of huh.Form objects that I have used as a solution back then. @bupd @NucleoFusion

pkg/views/base/listselect/model.go
```diff
@@ -0,0 +1,135 @@
...
+func NewModel(items []list.Item, construct string) Model {
+	const defaultWidth = 20
+	selected := make(map[int]struct{})
+	l := list.New(items, ItemDelegate{Selected: &selected}, defaultWidth, listHeight)
+	l.Title = "Select one or more " + construct + " (space to toggle, enter to confirm)"
+	l.SetShowStatusBar(true)
+	l.SetFilteringEnabled(true)
+	l.Styles.Title = views.TitleStyle
+	l.Styles.PaginationStyle = views.PaginationStyle
+	l.Styles.HelpStyle = views.HelpStyle
```
NucleoFusion, line 82: Help needs to have a "space" in the help menu

PR #714, opened 2026-02-18
feat(repo): Implement Repository Update Command

### LINKED ISSUE #710: Remove repository update cmd
## Summary 
We have an empty repository update cmd which does nothing - its better to remove that placeholder. 

https://github.com/goharbor/harbor-cli/blob/main/cmd/harbor/root/repository/update.go

bhavyaKhatri2703:
fixes #710 

This pr adds a repo update command which can update the description of a repository.


cmd/harbor/root/repository/update.go
```diff
@@ -12,3 +12,64 @@
...
+				}
+			} else {
+				projectName, err = prompt.GetProjectNameFromUser()
+				if err != nil {
+					return fmt.Errorf("failed to get project name: %w", err)
+				}
+				repoName = prompt.GetRepoNameFromUser(projectName)
+			}
+
+			if !cmd.Flags().Changed("description") {
```
qcserestipy: It would be great to make the flag not mandatory but optional. In harbor-cli we usually open a small user tui for entering the information needed. For example see other update commands.

pkg/api/repository_handler.go
```diff
@@ -107,3 +108,25 @@ func SearchRepository(query string) (search.SearchOK, error) {
...
 	return *response, nil
 }
+
+func UpdateRepository(projectName, repoName string, description string) error {
+	ctx, client, err := utils.ContextWithClient()
+	if err != nil {
+		return err
+	}
+
+	repo := &models.Repository{
```
qcserestipy, line 118: Here you are creating a new blank repository object without fetching the info of the already existing repo from the api. I would expect that this could lead to some issues. I think the correct behavior would be something like getting the repo object from the api, updating the description and then calling the update request to the api.

qcserestipy (changes requested):
Thank you for the contribution! I was wondering whether it makes sense to make `--description` flag optional and follow conventions for other update commands to open a view form where you can enter the description. Additionally, I guess it would be better to fetch the upstream repo api object first, update and then push the update instead of creating a new repo object without the other info. Could you also please add screenshots or a small screen record to demonstrate the working command.

NucleoFusion (approved):
lgtm!
Thanks for the contribution

qcserestipy (approved):
LGTM

bupd (approved):
/lgtm

PR #728, opened 2026-03-06
fix(ci): add temp container for git commit SHA in build

### LINKED ISSUE #726: Harbor CLI binary is missing git-commit SHA
## Summary

Harbor-CLI binary currently doesnt have git commit SHA included in its binary - we need to fix it, 
Context: previously it was present - problem is missing ldflags when building the binary, fix the dagger build binary to include the correct ldflags 

```
❯ harbor-cli version
INFO[0000] Data file at /home/bupd/.local/share/harbor-cli/data.yaml updated with new configPath: /home/bupd/.config/harbor-cli/config.yaml
Version:      0.0.18
Go version:   1.24.8
Git commit:
Built:        2026-03-03T14:43:23Z
OS/Arch:      linux/amd64
```

/label good-first-issue
/label bug

Sypher845:
Fixes #726 

* The `Build` function uses the Alpine version of the Go container, which does not have git installed.
* One option was to install git inside the build container, but that would run redundantly in the loop for all cross-platform builds.
* Instead, a temporary container is used outside the loop to resolve the git commit SHA once and reuse it for all builds.

image

### DIFF
--- a/.dagger/build.go
+++ b/.dagger/build.go
@@ -24,6 +24,14 @@ func (m *HarborCli) Build(ctx context.Context,
 	goos := []string{"linux", "darwin", "windows"}
 	goarch := []string{"amd64", "arm64"}
 
+	temp := dag.Container().
+		From("alpine:latest").
+		WithExec([]string{"apk", "add", "--no-cache", "git"}).
+		WithMountedDirectory("/src", source).
+		WithWorkdir("/src")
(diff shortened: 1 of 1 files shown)


bupd:
Looks like there is two gosec issues. Please find the root cause and fix it in a separate PR. 

Also we need to find why golangci-lint being latest fails the pipeline - we also need to find the cause for this thing

Issue #723, opened 2026-03-04
Discussion : Vulnerability Dashboard Command


Sypher845:
### What it does?

Harbor has a Security Hub that tracks vulnerabilities across all scanned artifacts. The vulnerability command will allow users to view vulnerability data from the terminal.

It uses the go-client's `Securityhub.ListVulnerabilities()` and `Securityhub.GetSecuritySummary()` methods.

## Proposed Approach
The following is a proposed structure for the command

- ###  Commands

| Command | What it does |
|---------|-------------|
| `harbor vulnerability summary` | Retrieve the vulnerability summary of the system |
| `harbor vulnerability list` | Lists individual vulnerabilities  |

- ### `summary` command and its flags

Shows a vulnerability overview across all artifacts.

| Flag | Description |
|------|-------------|
| `--with-dangerous-artifact` | Include top 5 most dangerous artifacts |
| `--with-dangerous-cve` | Include top 5 most dangerous CVEs |

- ###  `summary` command CLI output
The following output will appear when both flags are used. I am also experimenting with making the progress bar colorful.

```
$ harbor vuln summary --with-dangerous-artifact  --with-dangerous-cve

2 artifact(s), 2 scanned, 0 not scanned

Total Vulnerabilities: 360  (Fixable: 281)

  Critical   ███░░░░░░░░░░░░░░░░░░░░░░░░░░░   40
  High       ██████████░░░░░░░░░░░░░░░░░░░░  125
  Medium     █████████████░░░░░░░░░░░░░░░░░  161
  Low        ██░░░░░░░░░░░░░░░░░░░░░░░░░░░░   25
  Unknown    █░░░░░░░░░░░░░░░░░░░░░░░░░░░░░    9
  None       ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░    0

Top 5 Most Dangerous Artifacts
┌──────────────────────────────────────────────────────────────────────────────────────┐
│  Project…  Repository                Digest            Critical  High      Medium    │
│ ──────────────────────────────────────────────────────────────────────────────────── │
│  8         library/nginx             sha256:8269a7352  39        125       161       │
│  .         .                         .                 .         .         .         |
│  .         .                         .                 .         .         .         │
└──────────────────────────────────────────────────────────────────────────────────────┘

Top 5 Most Dangerous CVEs
┌────────────────────────────────────────────────────────────────────┐
│  CVE ID            Severity  CVSS3     Package           Version   │
[...]

PR #738, opened 2026-03-11
feat : vuln summary command

Sypher845:
Kooha-2026-03-15-18-42-39

> **Note** : This is the first PR for the issue #723 

## Description
The PR adds the new harbor `vuln summary command`. A new command that displays an overview of vulnerability data from Harbor's Security Hub.
- Severity distribution bar chart (Critical, High, Medium, Low, None, Unknown)
- Total artifact and scan counts with not-scanned breakdown
- Optional `--artifact` flag to show top 5 most vulnerable artifacts
- Optional `---cve` flag to show top 5 most dangerous CVEs

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

### DIFF
--- a/pkg/views/vulnerability/summary/view.go
+++ b/pkg/views/vulnerability/summary/view.go
@@ -15,12 +15,10 @@ package summary
 
 import (
 	"fmt"
-	"os"
 	"strconv"
 	"strings"
 
 	"github.com/charmbracelet/bubbles/table"
-	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/lipgloss"
 	"github.com/goharbor/go-client/pkg/sdk/v2.0/models"
 	"github.com/goharbor/harbor-cli/pkg/api"
@@ -36,27 +34,31 @@ func ViewVulnerabilitySummary(summary *models.SecuritySummary, showDangerousArti
 		return
 	}
 
-	fmt.Printf("%d artifact(s), %d scanned, %d not scanned\n\n", summary.TotalArtifact, summary.ScannedCnt, summary.TotalArtifact-summary.ScannedCnt)
+	var sb strings.Builder
 
-	fmt.Printf("Total Vulnerabilities: %s  (Fixable: %s)\n\n",
+	fmt.Fprintf(&sb, "%d artifact(s), %d scanned, %d not scanned\n\n", summary.TotalArtifact, summary.ScannedCnt, summary.TotalArtifact-summary.ScannedCnt)
+
+	fmt.Fprintf(&sb, "Total Vulnerabilities: %s  (Fixable: %s)\n\n",
 		views.BoldStyle.Render(strconv.FormatInt(summary.TotalVuls, 10)),
 		views.BoldStyle.Render(strconv.FormatInt(summary.FixableCnt, 10)),
 	)
 
-	renderBarChart(summary)
+	renderBarChart(&sb, summary)
 
 	if showDangerousArtifacts {
-		fmt.Println("Top 5 Most Dangerous Artifacts")
-		renderDangerousArtifacts(summary.DangerousArtifacts)
+		sb.WriteString("Top 5 Most Dangerous Artifacts\n")
+		renderDangerousArtifacts(&sb, summary.DangerousArtifacts)
 	}
 
 	if showDangerousCVEs {
-		fmt.Println("Top 5 Most Dangerous CVEs")
-		renderDangerousCVEs(summary.DangerousCves)
+		sb.WriteString("Top 5 Most Dangerous CVEs\n")
+		renderDangerousCVEs(&sb, summary.DangerousCves)
 	}
+
+	fmt.Print(sb.String())
 }
 
-func renderBarChart(summary *models.SecuritySummary) {
+func renderBarChart(sb *strings.Builder, summary *models.SecuritySummary) {
 	entries := []struct {
 		label string
 		count int64
@@ -82,12 +84,12 @@ func renderBarChart(summary *models.SecuritySummary) {
 		}
 		empty := barWidth - filled
 		bar := e.color.Render(strings.Repeat("█", filled)) + strings.Repeat("─", empty)
-		fmt.Printf("  %-10s %s %3.0f%% (%d CVEs)\n\n", e.label, bar, pct, e.count)
+		fmt.Fprintf(sb, "  %-10s %s %3.0f%% (%d CVEs)\n\n", e.label, bar, pct, e.count)
 	}
-	fmt.Println()
+	sb.WriteString("\n")
 }
 
-func renderDangerousArtifacts(artifacts []*models.DangerousArtifact) {
+func renderDangerousArtifacts(sb *strings.Builder, artifacts []*models.DangerousArtifact) {
 	columns := []table.Column{
 		{Title: "Project", Width: tablelist.WidthL},
 		{Title: "Repository", Width: tablelist.WidthXXL},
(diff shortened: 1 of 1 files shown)


pkg/views/vulnerability/summary/view.go
```diff
@@ -0,0 +1,164 @@
...
+		fmt.Println("No vulnerability summary data available.")
+		return
+	}
+
+	fmt.Printf("%d artifact(s), %d scanned, %d not scanned\n\n", summary.TotalArtifact, summary.ScannedCnt, summary.TotalArtifact-summary.ScannedCnt)
+
+	fmt.Printf("Total Vulnerabilities: %s%d%s  (Fixable: %s%d%s)\n\n",
+		views.BoldANSI, summary.TotalVuls, views.ResetANSI,
+		views.BoldANSI, summary.FixableCnt, views.ResetANSI,
+	)
```
NucleoFusion: Do not use ANSI styling like this. Instead rely on `lipgloss`.
have a look at the slog PR

cmd/harbor/root/vulnerability/summary.go
```diff
@@ -0,0 +1,66 @@
...
+				vulnsummary.ViewVulnerabilitySummary(summary, withDangerousArtifact, withDangerousCVE)
+			}
+
+			return nil
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.BoolVarP(&withDangerousArtifact, "with-dangerous-artifact", "", false, "Include top 5 dangerous artifact details in summary")
+	flags.BoolVarP(&withDangerousCVE, "with-dangerous-cve", "", false, "Include top 5 dangerous CVE details in summary")
```
NucleoFusion: Can we have either aliases or like short-forms of this?
I feel like writing `--with-dangerous-cve` and/or the artifact is just too big and tedious.
Something like 
`--cve & --art` or something for shortforms?
Or anything you can suggest

pkg/views/vulnerability/summary/view.go
```diff
@@ -0,0 +1,164 @@
...
+	for _, e := range entries {
+		filled := 0
+		if summary.TotalVuls > 0 {
+			filled = int(float64(e.count) / float64(summary.TotalVuls) * float64(barWidth))
+			if e.count > 0 && filled == 0 {
+				filled = 1
+			}
+		}
+		empty := barWidth - filled
+		bar := e.color + strings.Repeat("█", filled) + views.ResetANSI + strings.Repeat("░", empty)
```
NucleoFusion: I think we should replace the empty character. Although in mine it shows something like this

Image

And in most modern terminal emulator's they do the same. But for VMs or other cases they look how they do in the PR Description

```go
strings.Repeat("█", filled) + strings.Repeat("─", empty)
// ████████████────────
```
I think this is better. Cause the actual block is standard unicode and so is the -
The half-filled block looks crude to me (this may as well be a personal opinion)
@bupd @qcserestipy What do you think?

pkg/views/vulnerability/summary/view.go
```diff
@@ -0,0 +1,164 @@
...
+		fmt.Printf("  %-10s %s %4d\n\n", e.label, bar, e.count)
```
NucleoFusion: Can we also add a %age?
Like showing
```
[XXXXX-----] 43% (4444 CVEs)
```
Do you think this would be better?

NucleoFusion (changes requested):

cmd/harbor/root/vulnerability/summary.go
Sypher845: `--cve` and `--artifact`, is this good?

NucleoFusion: yeah, that would be good

pkg/views/vulnerability/summary/view.go
Sypher845: Yeah , adding it would be better as the UI is also having it.

pkg/views/vulnerability/summary/view.go
Sypher845: image\
image

i feel the dot one looks good

NucleoFusion: I feel the line one lol
But you can keep the dot, it isnt bad
@qcserestipy @bupd what are your thoughs?

pkg/views/vulnerability/summary/view.go
Sypher845: Waiting for #739 to get merged

bupd: its merged  - @Sypher845 - please update 

Thanks

pkg/views/vulnerability/summary/view.go
bupd: yeah the line looks good - feels more modern - and across the cli we use lines - so should be good to go..

bupd: 2 for 1 - I believe line is the way to go.

bupd (changes requested):
@Sypher845 please resolve conflicts

pkg/views/vulnerability/summary/view.go
Sypher845: Line wins!!

pkg/views/vulnerability/summary/view.go
```diff
@@ -0,0 +1,167 @@
...
+			strconv.FormatInt(artifact.MediumCnt, 10),
+		})
+	}
+
+	m := tablelist.NewModel(columns, rows, len(rows))
+
+	if _, err := tea.NewProgram(m).Run(); err != nil {
+		fmt.Println("Error running program:", err)
+		os.Exit(1)
+	}
```
NucleoFusion: I think it would be better to propagate the error.
Since with the addition of the errors package, we would then only need to handle error's in one place, the Root `cmd.Execute()`.

pkg/views/vulnerability/summary/view.go
```diff
@@ -0,0 +1,167 @@
...
+			cve.Version,
+		})
+	}
+
+	m := tablelist.NewModel(columns, rows, len(rows))
+
+	if _, err := tea.NewProgram(m).Run(); err != nil {
+		fmt.Println("Error running program:", err)
+		os.Exit(1)
+	}
```
NucleoFusion: I think it would be better to propagate the error.
Since with the addition of the errors package, we would then only need to handle error's in one place, the Root `cmd.Execute()`.

pkg/views/vulnerability/summary/view.go
```diff
@@ -0,0 +1,167 @@
...
+	if showDangerousArtifacts {
+		fmt.Println("Top 5 Most Dangerous Artifacts")
+		renderDangerousArtifacts(summary.DangerousArtifacts)
+	}
+
+	if showDangerousCVEs {
+		fmt.Println("Top 5 Most Dangerous CVEs")
+		renderDangerousCVEs(summary.DangerousCves)
+	}
+}
```
NucleoFusion, line 59: Also I think doing something like
`strings.Builder`
and then appending everything would be better, to avoid the delay we have between the barchart renders and the top5 renders

Sypher845: the top5 tables are using `Bubbletea`, can we apppend them?, as they run seperately , and do not have plain fmt.Println

NucleoFusion: You can do it like this
```go
	m := tablelist.NewModel(columns, rows, len(rows))
	fmt.Println(m.View())
```

Just don't `Run()` it.

pkg/views/vulnerability/summary/view.go
Sypher845: so if we dont run it , there will be no error to propagate

NucleoFusion: yep, so you can ignore this

NucleoFusion: resolve these issues if you can
like mark as resolved

PR #748, opened 2026-03-16
feat/vuln list command

Sypher845:
> **Note:** This is the second PR for the issue #723 

## Description
The PR adds the new `harbor vuln list` command. A command  that displays the list of vulnerabilities along with filtering from the Security Hub.

image

## Command usage
-  `harbor vuln list ` (No filters)
Kooha-2026-03-17-04-14-39
The CVE-IDs are hyperlinked  to aquasec website (provided by the api)

-  `harbor vuln list --query "k=v , , , k=[min~max]"` : Using query flag based filtering
Kooha-2026-03-17-04-22-05

-  `harbor vuln list (flags based filtering)` : Using flags for filtering 
Kooha-2026-03-17-04-27-36

-  `harbor vuln list --fixable`: Show only fixable vulnerabilities (cli side)
Kooha-2026-03-17-04-29-26

-  `harbor vuln list --exclude` : woks oppposite of query flag (cli side)
Kooha-2026-03-17-04-37-00

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

### DIFF
--- a/cmd/harbor/root/vulnerability/list.go
+++ b/cmd/harbor/root/vulnerability/list.go
@@ -16,6 +16,7 @@ package vulnerability
 import (
 	"fmt"
 
+	"github.com/goharbor/go-client/pkg/sdk/v2.0/models"
 	"github.com/goharbor/harbor-cli/pkg/api"
 	"github.com/goharbor/harbor-cli/pkg/utils"
 	vulnlist "github.com/goharbor/harbor-cli/pkg/views/vulnerability/list"
@@ -42,24 +43,24 @@ func ListVulnerabilitiesCommand() *cobra.Command {
 				return fmt.Errorf("page size should be less than or equal to 100")
 			}
 
-			response, err := api.ListVulnerabilities(opts)
+			allVulnerabilities, hasNext, err := fetchVulnerabilities(opts)
 			if err != nil {
 				return fmt.Errorf("failed to list vulnerabilities: %v", utils.ParseHarborErrorMsg(err))
 			}
 
-			if len(response.Payload) == 0 {
+			if len(allVulnerabilities) == 0 {
 				log.Info("No vulnerabilities found")
 				return nil
 			}
 
 			formatFlag := viper.GetString("output-format")
 			if formatFlag != "" {
-				err = utils.PrintFormat(response.Payload, formatFlag)
+				err = utils.PrintFormat(allVulnerabilities, formatFlag)
 				if err != nil {
 					return err
 				}
 			} else {
-				vulnlist.ViewVulnerabilityList(response.Payload)
+				vulnlist.ViewVulnerabilityList(allVulnerabilities, hasNext)
 			}
 
 			return nil
@@ -71,7 +72,7 @@ func ListVulnerabilitiesCommand() *cobra.Command {
 	flags.Int64VarP(&opts.PageSize, "page-size", "", 10, "Size of per page")
 	flags.StringVarP(&opts.Q, "query", "q", "", "Filter vulnerabilities with a ',' separated query string like exact k=v and range k=[min~max]")
 	flags.StringVarP(&opts.CVEID, "cve-id", "", "", "Filter by exact CVE ID")
-	flags.StringVarP(&opts.CVSSScore, "cvss-score", "", "", "Filter by CVSS v3 score range (e.g. [7.0~10.0])")
+	flags.StringVarP(&opts.CVSSScore, "cvss-score", "", "", "Filter by CVSS v3 score range (e.g. 7.0~10.0) or exact score (e.g. 7.0)")
 	flags.StringVarP(&opts.Severity, "severity", "", "", "Filter by severity level")
 	flags.StringVarP(&opts.Repository, "repository", "", "", "Filter by exact repository name")
 	flags.StringVarP(&opts.ProjectName, "project-name", "", "", "Filter by exact project name")
@@ -82,4 +83,34 @@ func ListVulnerabilitiesCommand() *cobra.Command {
 	flags.BoolVarP(&opts.Fixable, "fixable", "", false, "Only show fixable vulnerabilities")
 
 	return cmd
+}
+
+func fetchVulnerabilities(opts api.ListVulnerabilityOptions) ([]*models.VulnerabilityItem, bool, error) {
+	var allVuln []*models.VulnerabilityItem
+	if opts.PageSize == 0 {
+		log.Debug("Page size is 0, will fetch all vulnerabilities")
+		opts.PageSize = 100
+		opts.Page = 1
+		for {
+			response, err := api.ListVulnerabilities(opts)
+			if err != nil {
+				return nil, false, fmt.Errorf("failed to list vulnerabilities: %v", utils.ParseHarborErrorMsg(err))
+			}
+			if len(response.Payload) == 0 {
+				break
+			}
+			allVuln = append(allVuln, response.Payload...)
+			opts.Page++
+			if opts.Page > 10 {
+				return allVuln, true, nil
+			}
+		}
+	} else {
+		response, err := api.ListVulnerabilities(opts)
+		if err != nil {
+			return nil, false, fmt.Errorf("failed to list vulnerabilities: %v", utils.ParseHarborErrorMsg(err))
+		}
+		allVuln = append(allVuln, response.Payload...)
+	}
+	return allVuln, false, nil
 }
--- a/cmd/harbor/root/vulnerability/list.go
+++ b/cmd/harbor/root/vulnerability/list.go
@@ -35,12 +35,8 @@ func ListVulnerabilitiesCommand() *cobra.Command {
 		Example: `  harbor vulnerability list`,
 		Args:    cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if opts.PageSize < 0 {
-				return fmt.Errorf("page size must be greater than or equal to 0")
-			}
-
-			if opts.PageSize > 100 {
-				return fmt.Errorf("page size should be less than or equal to 100")
+			if opts.PageSize <= 0 || opts.PageSize > 100 {
+				return fmt.Errorf("page size must be greater than 0 and less than or equal to 100")
 			}
 
 			allVulnerabilities, hasNext, err := fetchVulnerabilities(opts)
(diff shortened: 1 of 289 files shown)


pkg/api/vulnerability_handler.go
```diff
@@ -34,3 +39,105 @@ func GetVulnerabilitySummary(withDangerousArtifact, withDangerousCVE bool) (*sec
...
+}
+
+func buildVulnerabilityQuery(opts ListVulnerabilityOptions) (string, error) {
+	var queries []string
+	if opts.CVEID != "" {
+		queries = append(queries, fmt.Sprintf("cve_id=%s", opts.CVEID))
+	}
+	if opts.CVSSScore != "" {
+		queries = append(queries, fmt.Sprintf("cvss_score_v3=%s", opts.CVSSScore))
+	}
```
NucleoFusion, line 175: CVSS Score is a range match, not an exact match as per the docs
```
cve_id(exact match)
cvss_score_v3(range condition)
severity(exact match)
status(exact match)
repository_name(exact match)
project_id(exact match)
package(exact match)
tag(exact match)
digest(exact match)
```

This is the format `k3=[min~max]`

pkg/api/vulnerability_handler.go
```diff
@@ -34,3 +39,105 @@ func GetVulnerabilitySummary(withDangerousArtifact, withDangerousCVE bool) (*sec
...
+	if opts.Severity != "" {
+		queries = append(queries, fmt.Sprintf("severity=%s", opts.Severity))
+	}
```
NucleoFusion, line 181: Check if severity is right or not, because severity can only be a handful of strings.
And harbor wont return an error, it will just return empty array.

pkg/api/vulnerability_handler.go
```diff
@@ -34,3 +39,105 @@ func GetVulnerabilitySummary(withDangerousArtifact, withDangerousCVE bool) (*sec
...
+	if opts.Tag != "" {
+		queries = append(queries, fmt.Sprintf("tag=%s", opts.Tag))
+	}
+	if opts.Digest != "" {
+		queries = append(queries, fmt.Sprintf("digest=%s", opts.Digest))
+	}
+	if opts.Q != "" {
+		queries = append(queries, opts.Q)
+	}
```
NucleoFusion: Add a FIXME here, 
And reference this in my draft PR #731.
This needs to be in the standard query format, but I plan to make changes to that, so you need not do it rn. But make a reference in the codebase (a FIXME comment saying what) and in the PR

NucleoFusion (changes requested):

pkg/api/vulnerability_handler.go
Sypher845: so if the users inputs just a particular number , we convret it to [val~val], is this the thing wanted?

NucleoFusion: No, I would find out a way to make a range. maybe have the flag be a string and tell in the `help` that the format is `min:max` or you can think of something better.
If we _can_ have range, we _should_ have range

Sypher845: Currently i have implemented the string one , `[min~max]`, will look for something better

Sypher845: @NucleoFusion  taking input in the form `[min~max]` is good ?, or we go with just  `min:max` , or `min~max`

NucleoFusion: I think either of the other. I think min:max cause that is what the query params support as well?
Double check that tho.

The [] seems redundant

Sypher845: The query params support `[min~max]` , so i guess we can take input in `min:max`, and then convert it in the required format for query, does this works?

NucleoFusion: Yes, I meant how the --range flag in project list command. Same syntax that it has

cmd/harbor/root/vulnerability/list.go
```diff
@@ -0,0 +1,116 @@
...
+		Example: `  harbor vulnerability list`,
+		Args:    cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if opts.PageSize < 0 {
+				return fmt.Errorf("page size must be greater than or equal to 0")
+			}
+
+			if opts.PageSize > 100 {
+				return fmt.Errorf("page size should be less than or equal to 100")
+			}
```
NucleoFusion: Could be joined in one with ||

cmd/harbor/root/vulnerability/list.go
```diff
@@ -0,0 +1,116 @@
...
+				return nil, false, fmt.Errorf("failed to list vulnerabilities: %v", utils.ParseHarborErrorMsg(err))
+			}
+			if len(response.Payload) == 0 {
+				break
+			}
+			allVuln = append(allVuln, response.Payload...)
+			opts.Page++
+			if opts.Page > 10 {
+				return allVuln, true, nil
+			}
```
NucleoFusion: you say we fetch all vulnerabilities but fetch only a thousand. Which are we supposed to do?

NucleoFusion (changes requested):
Instead of `--page-size 0` I think a special flag would be better, since its not that user friendly and hard to describe in the help

cmd/harbor/root/vulnerability/list.go
Sypher845: As discussed in meet, we will show till 1000, and then provide a link to UI
As showing so much vuln in cli, would not be good

Sypher845:
`--all` flag works?

cmd/harbor/root/vulnerability/list.go
Sypher845: @NucleoFusion can we do something like
`flags.BoolVarP(&opts.All, "all", "", false, "Show all vulnerabilities (up to 1000)")`
or something else can be done?

NucleoFusion: yes this is better

cmd/harbor/root/vulnerability/list.go
```diff
@@ -0,0 +1,113 @@
...
+	flags.Int64VarP(&opts.PageSize, "page-size", "", 10, "Size of per page")
+	flags.StringVarP(&opts.Q, "query", "q", "", "Filter vulnerabilities with a ',' separated query string like exact k=v and range k=[min~max]")
+	flags.StringVarP(&opts.CVEID, "cve-id", "", "", "Filter by exact CVE ID")
+	flags.StringVarP(&opts.CVSSScore, "cvss-score", "", "", "Filter by CVSS v3 score range (e.g. 7.0~10.0) or exact score (e.g. 7.0)")
+	flags.StringVarP(&opts.Severity, "severity", "", "", "Filter by severity level")
+	flags.StringVarP(&opts.Repository, "repository", "", "", "Filter by exact repository name")
+	flags.StringVarP(&opts.ProjectName, "project-name", "", "", "Filter by exact project name")
+	flags.StringVarP(&opts.Package, "package", "", "", "Filter by exact package name")
+	flags.StringVarP(&opts.Tag, "tag", "", "", "Filter by exact artifact tag")
+	flags.StringVarP(&opts.Digest, "digest", "", "", "Filter by exact artifact digest")
```
NucleoFusion, line 80: @bupd @qcserestipy 
Should we have this many flags? would it be better to just have the `Q` param?
And since I am planning a PR on adding `validKeys` and usage to the query param about how it works in #731 .

Cause managing the current `BuildQuery` function takes the flags and just creates the full query, so integrating these two would be a little unintuitive.
And then also another problem is the individual flags are rigid, no support for exact/fuzzy switch, and there are also the Q params rigidity, like the params dont allow/follow 2 exact matching, they only do one. Using the Q would also allow `or` and `and`.

NucleoFusion (commented):
this look good, but waiting for some clarifications

cmd/harbor/root/vulnerability/list.go
Sypher845: one thing to note, the SecurityHub API doesn't actually suport the fuzzy match, union and intersection for vuln fields.

pkg/api/vulnerability_handler.go
```diff
@@ -34,3 +39,186 @@ func GetVulnerabilitySummary(withDangerousArtifact, withDangerousCVE bool) (*sec
...
+			"Critical": true,
+			"High":     true,
+			"Medium":   true,
+			"Low":      true,
+			"n/a":      true,
+			"None":     true,
+		}
+		if !allowedSeverities[opts.Severity] {
+			return "", fmt.Errorf("invalid severity value: %s. Allowed values are: Low, Medium, High, Critical, n/a, None", opts.Severity)
+		}
```
NucleoFusion: This should be done in a separate helper function, and use a switch case instead of allocating memory for a hashmap we use once.

pkg/api/vulnerability_handler.go
```diff
@@ -34,3 +39,186 @@ func GetVulnerabilitySummary(withDangerousArtifact, withDangerousCVE bool) (*sec
...
+	if opts.Digest != "" {
+		queries = append(queries, fmt.Sprintf("digest=%s", opts.Digest))
+	}
+
+	return strings.Join(queries, ","), nil
+}
+
+func parseVulnerabilityExcludeMap(exclude string) map[string]string {
+	excludeMap := make(map[string]string)
+	for _, query := range strings.Split(exclude, ",") {
```
NucleoFusion, line 216: Doesnt `splitN` always return 2 items? Well, _at most_ 2 items, it wouldnt go beyond 2 anyway.
Better would be normal Split and then checking if its 2.

pkg/api/vulnerability_handler.go
```diff
@@ -34,3 +39,186 @@ func GetVulnerabilitySummary(withDangerousArtifact, withDangerousCVE bool) (*sec
...
+			}
+		}
+		if len(res) == 0 {
+			return &securityhub.ListVulnerabilitiesOK{
+				Payload: nil,
+			}, nil
+		}
+		return &securityhub.ListVulnerabilitiesOK{
+			Payload: res,
+		}, nil
```
NucleoFusion: You probably should not create these types yourself since these are swagger generated.
It would make much more sense if we had something like
`[]VulnItem` instead of the whole type.

NucleoFusion (changes requested):
minor changes, works well

NucleoFusion (approved):
lgtm!
Thanks for the contribution

bupd (approved):
/lgtm

PR #754, opened 2026-03-18
fix: allow admin users to skip old password input

### LINKED ISSUE #743: [bug]: Password change command only supports system admin
## Description
The current implementation of the password command only sends the `newPassword` in request. As per the swagger API,
```
Password to be updated, the attribute 'old_password' is optional when the API is called by the system administrator
```
We need to add support for the oldPassword field and/or an --admin flag, or check if the user is an admin at command start.

## Steps to Reproduce
1. Run command: `harbor-cli user password` as a normal user
2. Observe error

## Expected Behavior
Change the password

## Actual Behavior

Error 403
```
ERRO[0013] Permission denied: Admin privileges are required to execute this command.
```

adityashekhawat1999:
## Description
This updates the `user password` command to allow admin users to skip entering the old password.

Right now, the CLI always requires the old password but according to the API behaviour it’s optional for admins.

- Fixes #743 

## Type of Change

- [ ✅] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
- Added `--admin` flag to the password command
- Updated validation to allow empty old password when `--admin` is used

### DIFF
--- a/doc/cli-docs/harbor-password.md
+++ b/doc/cli-docs/harbor-password.md
@@ -15,8 +15,8 @@ harbor password [flags]
 ### Options
 
 ```sh
-      --admin  admin password change
-  -h, --help   help for password
+      --admin   Admin password change
+  -h, --help    help for password
 ```
 
 ### Options inherited from parent commands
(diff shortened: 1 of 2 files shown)


adityashekhawat1999:
@NucleoFusion The cli would still send the old password field(empty for admin), i wanted to ask is that acceptable for the API or should the field be omitted entirely

PR #780, opened 2026-03-31
refac: `artifact list` command with `errors` package

### LINKED ISSUE #635: Proposal: Fix improper logging and error handling
# Overview

> **This is a proposal to update logger and fix error handling in general.**

## Origin

I have been using `harbor` CLI to manage my local registry for a while. So when I was trying to visit the audit logs of my registry using `harbor logs`, the process exited with a status code **1** without any proper logging or reason.

Image

Then after some time I found that the registry hasn't been started using `harbor health` (*That's my bad*), but though there isn't any proper response from the tool. So I just started to test other commands and their sub-commands, where I get varied response and logging. 

Surprisingly upon executing `harbor user elevate`, The program just panicked. 

Image

Also even the `harbor user create` just gets all the inputs and exits without proper logging, irrespective of the status of the registry.

Image

## More Context

When I was working on this #634, the logger didn't log anything, as the logrus logger didn't set the output writer properly. 
This may be a quick fix, but there are lot of lines of code performing logging, scattered across files.

Additionally, the logrus logger library which is being used in the project, has been in **maintenance mode for nearly 6 years**. ([ref](https://github.com/sirupsen/logrus/commit/4ecd9a62bd57b6533e0d87ae23a6b10adc32b5ce)) 

Due to these, the errors neither handled properly nor being logged. And at some places, errors are simply ignored, which isn't a good practice. ([example](https://github.com/goharbor/harbor-cli/blob/main/pkg/prompt/prompt.go#L189)).

# Proposal

With all the above considerations in mind, I suggest to,
1. Use the Charm's [`log`](https://github.com/charmbracelet/log?tab=readme-ov-file#log) library. 
  > *I suggest using the above lib, as the TUI is developed using Charm's components, styles and will be easy to set the theme across the CLI entirely*

2. Build a logger package using **Cobra CLI's** builtin `stdout` & `stderr` writter. 
3. Or use of any other logger such as [Zerolog](https://github.com/rs/zerolog), [Zap](https://github.com/uber-go/zap).

And for errors, I would highly suggest to define an internal `errors` package listing and mapping all the possible errors, that the tool may encounter. Like how it is implemented properly in the Harbor and its `go-client` SDK itself. ([example](https://github.com/goharbor/harbor/tree/main/src/lib/errors))

**So request the maintainers and the community to go through this proposal and share your thoughts and recommendations in this issue thread.**

cc: @bupd @Vad1mo @OrlinVasilev @qcserestipy @NucleoFusion

vg006:
## Description
This PR demonstrates the implementation of the project's `pkg/errors` package in the #636, by replacing the existing non-uniform error return statements with the `pkg/errors` package APIs.

- Fixes #635 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
- Replace the return statements with `fmt.Errorf`, `errors.New` with the built-in `pkg/errors`
### Overview
e1
e2
e3

### DIFF
--- a/cmd/harbor/root/artifact/list.go
+++ b/cmd/harbor/root/artifact/list.go
@@ -56,12 +56,12 @@ Supports pagination, search queries, and sorting using flags.`,
 			if len(args) > 0 {
 				projectName, repoName, err = utils.ParseProjectRepo(args[0])
 				if err != nil {
-					return errors.New("Invalid project/repository format", "expected format: <project>/<repository>")
+					return err
 				}
 			} else {
 				projectName, err = prompt.GetProjectNameFromUser()
 				if err != nil {
-					return errors.AsError(err).WithMessage("failed to get project name")
+					return err
 				}
 				repoName = prompt.GetRepoNameFromUser(projectName)
 			}
@@ -69,7 +69,7 @@ Supports pagination, search queries, and sorting using flags.`,
 			artifacts, err = api.ListArtifact(projectName, repoName, opts)
 
 			if err != nil {
-				return errors.AsError(err).WithMessage("failed to list artifacts")
+				return err
 			}
 
 			FormatFlag := viper.GetString("output-format")
--- a/cmd/harbor/root/artifact/list.go
+++ b/cmd/harbor/root/artifact/list.go
@@ -14,10 +14,9 @@
 package artifact
 
 import (
-	"fmt"
-
 	"github.com/goharbor/go-client/pkg/sdk/v2.0/client/artifact"
 	"github.com/goharbor/harbor-cli/pkg/api"
+	"github.com/goharbor/harbor-cli/pkg/errors"
 	"github.com/goharbor/harbor-cli/pkg/prompt"
 	"github.com/goharbor/harbor-cli/pkg/utils"
 	artifactViews "github.com/goharbor/harbor-cli/pkg/views/artifact/list"
@@ -44,11 +43,11 @@ Supports pagination, search queries, and sorting using flags.`,
 		Args: cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if opts.PageSize < 0 {
-				return fmt.Errorf("page size must be greater than or equal to 0")
+				return errors.New("Invalid page size", "page size must be greater than or equal to 0")
 			}
 
 			if opts.PageSize > 100 {
-				return fmt.Errorf("page size should be less than or equal to 100")
+				return errors.New("Invalid page size", "page size must be less than or equal to 100")
 			}
 			var err error
 			var artifacts artifact.ListArtifactsOK
@@ -57,12 +56,12 @@ Supports pagination, search queries, and sorting using flags.`,
 			if len(args) > 0 {
 				projectName, repoName, err = utils.ParseProjectRepo(args[0])
 				if err != nil {
-					return fmt.Errorf("failed to parse project/repo: %v", err)
+					return errors.New("Invalid project/repository format", "expected format: <project>/<repository>")
 				}
 			} else {
 				projectName, err = prompt.GetProjectNameFromUser()
 				if err != nil {
-					return fmt.Errorf("failed to get project name: %v", utils.ParseHarborErrorMsg(err))
+					return errors.AsError(err).WithMessage("failed to get project name")
 				}
 				repoName = prompt.GetRepoNameFromUser(projectName)
 			}
@@ -70,13 +69,12 @@ Supports pagination, search queries, and sorting using flags.`,
 			artifacts, err = api.ListArtifact(projectName, repoName, opts)
 
 			if err != nil {
-				return fmt.Errorf("failed to list artifacts: %v", err)
+				return errors.AsError(err).WithMessage("failed to list artifacts")
 			}
 
 			FormatFlag := viper.GetString("output-format")
 			if FormatFlag != "" {
-				err = utils.PrintFormat(artifacts, FormatFlag)
-				if err != nil {
+				if err = utils.PrintFormat(artifacts, FormatFlag); err != nil {
 					return err
 				}
 			} else {
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -22,23 +22,96 @@ var (
 	as = errors.As
 )
 
+type Frame struct {
+	Message string
+	Hints   []string
+}
+
 type Error struct {
-	cause   error
-	message string
-	hints   []string
+	frames []Frame
+	cause  error
 }
 
 func New(message string) *Error {
 	return &Error{
-		message: message,
+		frames: []Frame{{Message: message}},
 	}
 }
 
 func Newf(format string, args ...any) *Error {
 	return &Error{
-		message: fmt.Sprintf(format, args...),
+		frames: []Frame{{Message: fmt.Sprintf(format, args...)}},
 	}
 }
+
+func Wrap(err error, message string) *Error {
+	e := AsError(err)
+	e.frames = append([]Frame{{Message: message}}, e.frames...)
+	return e
+}
+
+func Wrapf(err error, format string, args ...any) *Error {
+	return Wrap(err, fmt.Sprintf(format, args...))
+}
+
+func (e *Error) WithHint(hint string) *Error {
+	if len(e.frames) > 0 {
+		e.frames[0].Hints = append(e.frames[0].Hints, hint)
+	}
+	return e
+}
+
+func (e *Error) WithCause(cause error) *Error {
+	if e.cause == nil {
+		e.cause = cause
+	}
+	return e
+}
+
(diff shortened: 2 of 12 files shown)


cmd/harbor/root/artifact/list.go
```diff
@@ -57,26 +56,25 @@ Supports pagination, search queries, and sorting using flags.`,
 			if len(args) > 0 {
 				projectName, repoName, err = utils.ParseProjectRepo(args[0])
 				if err != nil {
-					return fmt.Errorf("failed to parse project/repo: %v", err)
+					return errors.New("Invalid project/repository format", "expected format: <project>/<repository>")
```
NucleoFusion: This and other instances of this are wrong no? since we would use the `WithHint` or `WithMessage` here instead of a new error?

vg006: Actually, the APIs are changed a bit. `WithHint` is removed and merged with `WithMessage` which accepts the hints optionally.

In the previous style, at different layers, we could add both **messages and hints optionally**. So this caused issues in rendering the tree view. 

Thus, I updated the `WithMessage` method to accept the hints optionally, but the messages aren't optional. So hints are passed only if a message is conveyed at that level.

vg006: You can check the signature of `WithMessage` [here](https://github.com/goharbor/harbor-cli/pull/780/changes#diff-ab01ccb939ab896a48c5445d1d755bf861ad2f337d98013132834c6d88641c8aR58)

NucleoFusion: yeah but shouldn't it still chain here, instead of returning a new one?

vg006: That's my bad. Thanks for noting it down.

Actually for initial testing, I refactored simple errors in the `artifact/list.go` and this is one of such. After I refactored all returns functions that are invoked at each layer. So I forgot to replace the initial one with the actually refactored one at the primitive layer. I have updated it now.

vg006: You can view the related changes here [[1](https://github.com/goharbor/harbor-cli/pull/780/changes#diff-503c0004fcbeef738432f5ca0588a2c132f0d1134d80552442da46858cebc417R59)][[2](https://github.com/goharbor/harbor-cli/pull/780/changes#diff-81ddbadfb415ccbb9c7af84f11668d1aa5e53c34025bf86d4702f16b4e42f045R69)]

pkg/errors/errors.go
```diff
@@ -0,0 +1,213 @@
...
+var (
+	as = errors.As
+)
+
+type Frame struct {
+	Message string
+	Hints   []string
+}
+
+type Error struct {
```
qcserestipy, line 34: I really like this package and think it is great. I wanted to note that it might be worth to document that we are shadowing the std errors package and explicitely telling people that they have to be careful about it (Like making a dedicated docs md for this).
A good way to let people do this difference could be this:
```go
import (
    stderrors "errors"
    "github.com/goharbor/harbor-cli/pkg/errors"
)
```

pkg/views/styles.go
```diff
@@ -37,10 +35,24 @@ var (
...
+
+	ErrHintStyle = lipgloss.NewStyle().
+			Foreground(lipgloss.Color("44"))
+
+	ErrEnumeratorStyle = lipgloss.NewStyle().
+				Foreground(lipgloss.Color("242")).MarginRight(1)
+)
+
+const (
+	GreenANSI = "\033[32m"
```
qcserestipy, line 54: To me it looks like that those colors and styles are already defined further up. Also, there are not used anywhere.

pkg/errors/errors.go
```diff
@@ -0,0 +1,213 @@
...
+			EnumeratorStyle(views.ErrEnumeratorStyle).
+			ItemStyle(views.ErrHintStyle)
+
+		for _, h := range rootFrame.Hints {
+			hintsTree.Child(h)
+		}
+		parts = append(parts, hintsTree.String())
+	}
+
+	if len(e.frames) > 1 {
```
qcserestipy, line 116: It would be great if you could test whether the tree structure output breaks in pipeline environments. I do not know whether all terms or envs would support this or expect single line error messages. Maybe @bupd and @NucleoFusion have some ideas, too.

qcserestipy (commented):
Thanks for the contribution! Looks very cool. I just left a few comments here and there.

pkg/errors/errors.go
vg006: Thank you for supporting my proposal. Yeah, maintaining proper documentation for this package will help maintainers and contributors to refactor either the project or the package itself. I think adding that doc to the `CONTRIBUTING.md` will inform the new contributors at the first place, if they work on this package or try to handle errors.

Also yeah, this package wraps the std `errors`, enforces to use it rather for uniformity across the codebase. Thus we could instruct the contributors to use this package instead of the std. So there will be only,
```go
import (
    "github.com/goharbor/harbor-cli/pkg/errors"
)
```

pkg/views/styles.go
vg006: Yeah the styles that are declared using the `lipgloss` library are the ones that I introduced newly. But the hardcoded ANSI sequences were defined earlier in the project itself and removed by @NucleoFusion last month ([ref](https://github.com/goharbor/harbor-cli/commit/eb788c055c526f68b1ba3a093b2799f1564f82ef#diff-950caf65f8ad078e4d5f15b0ea7d2e422a24aa09ffa4b060624b619a5f9e4ca7L42-L47)). I don't know why, but I suspect an issue while rebasing. I will check it and clean it up.

pkg/errors/errors.go
vg006: This tree output is planned to only be returned on non-verbose mode, *where it's used by a real user*. But for the CI and any other environment, the verbose mode of logging will be used I guess. So this tree won't be renderded, rather the error will be detailed by the logger as in the PR #722.

Even if CI or other envs use non-verbose mode, it will work I believe, because the ANSI sequences are valid UTF-8 encoded, won't disrupt tools like `grep`, `rg` parsing them, *when piped the harbor output in their scripts.*

For example, I tested the following in the local env to check compatibility with `grep`
```go
package main

import (
	"fmt"

	"github.com/charmbracelet/lipgloss"
)

func main() {
	style := lipgloss.NewStyle().
		Bold(true).
		Foreground(lipgloss.Color("205")).
		Padding(1)
	output := style.Render("Hello, Pipeline!")
	fmt.Println(output)
}
```

### Preview
ci

vg006:
Thank you @qcserestipy for supporting and reviewing the package. However, the above comments are very specific to the `errors` package that we are developing in the PR #636, *as this PR is a sample, demonstrating the `errors` package by refactoring the `artifact list` command.*

As you really interested in the betterment of this package, I really appreciate you to look into it PR #636 and will be happy to discuss further in there, *will be helpful to all track relevant reviews and changes properly, that's why*.

Thank you.

NucleoFusion:
> Thank you @qcserestipy for supporting and reviewing the package. However, the above comments are very specific to the `errors` package that we are developing in the PR #636, _as this PR is a sample, demonstrating the `errors` package by refactoring the `artifact list` command._
> 
> As you really interested in the betterment of this package, I really appreciate you to look into it PR #636 and will be happy to discuss further in there, _will be helpful to all track relevant reviews and changes properly, that's why_.
> 
> Thank you.

Isnt this PR much suited for merging? since we also get a reference material which the umbrella issue can mention and even other contributors can pick up on?

I think the changes (aside from the errors package) are simple enough that we wont need to spend a lot of extra time reviewing it.
@qcserestipy what do you think?

vg006:
Yeah @NucleoFusion,
- This PR is *partially* ready to merge, as it needs some more subtle changes.
- This could be used as a referrence and could create an umbrella issue, tracking all the migrations.
- Also the migration is quite simple and easier to review and manage, if we move command-by-command.

pkg/views/styles.go
vg006: Hello @qcserestipy, I updated the file and now it is compatible with the upstream, you may review it [here](https://github.com/goharbor/harbor-cli/pull/636/changes/c77313eb737c06b7adc3912b02685bcb47010141#diff-950caf65f8ad078e4d5f15b0ea7d2e422a24aa09ffa4b060624b619a5f9e4ca7). Thank you.

pkg/errors/errors.go
NucleoFusion: No no, as of now, the logger never logs errors, they get transported back to the cobra root. Where they are handled _by_ the errors package.

with `RunE` the errors are handled here, i believe,
```
func main() {
	err := root.RootCmd().Execute()
	if err != nil {
		os.Exit(1)
	}
}
```

And right now, i dont know how we would be able to distinguish between the verbose and non-verbose mode since we are outside of the command.

PR #785, opened 2026-03-31
feat: allow localhost as a valid host in URL validation

### LINKED ISSUE #783: Allow localhost as valid server address in login
localhost is not working as valid server address. We need to fix the regex regarding this. 

Image

marwan562:
## Description
This pull request modifies the `ValidateURL` function to allow `localhost` as a valid host. Previously, local development addresses lacking a top-level domain (like `localhost` or `http://localhost:8080`) failed the domain validation regex, which prevented users from successfully logging into local Harbor instances.

- Fixes #783

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
- Added an early return check for `"localhost"` in `pkg/utils/helper.go`'s `ValidateURL` function before the strict domain regex is evaluated.
- Expanded the URL validation test suite in `pkg/utils/helper_test.go` to include coverage for `localhost` addresses using both HTTP and HTTPS schemes, as well as with and without ports.

### DIFF
--- a/pkg/utils/helper.go
+++ b/pkg/utils/helper.go
@@ -196,6 +196,10 @@ func ValidateURL(rawURL string) error {
 		return nil
 	}
 
+	if host == "localhost" {
+		return nil
+	}
+
 	if !domainNameRegex.MatchString(host) {
 		return fmt.Errorf("invalid host: must be a valid IP address or domain name")
 	}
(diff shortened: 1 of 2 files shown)


NucleoFusion (approved):
lgtm
Thanks for the contribution

bupd (approved):
/lgtm

PR #802, opened 2026-04-09
feat: immutable tags update subcommand

NishchayRajput:
## Description
The Harbor CLI currently lacks support for updating existing immutable rules via CLI. 
`harbor tag immutable update [PROJECT_NAME]` command with interactive-first behavior if project name not mentioned.

- Fixes #798 

## Type of Change

```bash
harbor tag immutable update [PROJECT_NAME]
  --rule-id <id>           # Rule ID to update (optional; interactive if omitted)
  --tag-pattern <pattern>  # New tag matching pattern (optional; interactive if omitted)
  --scope <scope>          # Scope (include/exclude) (optional; interactive if omitted)
  --file <path>            # Load rule update config from JSON file (optional)
```

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
- cmd/harbor/root/tag/immutable/cmd.go
- cmd/harbor/root/tag/immutable/update.go
- pkg/api/immutable_handler.go

### DIFF
--- a/cmd/harbor/root/tag/immutable/update.go
+++ b/cmd/harbor/root/tag/immutable/update.go
@@ -1,0 +1,169 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package immutable
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/goharbor/go-client/pkg/sdk/v2.0/models"
+	"github.com/goharbor/harbor-cli/pkg/api"
+	"github.com/goharbor/harbor-cli/pkg/prompt"
+	"github.com/goharbor/harbor-cli/pkg/utils"
+	"github.com/goharbor/harbor-cli/pkg/views/immutable/create"
+	"github.com/spf13/cobra"
+)
+
+func UpdateImmutableCommand() *cobra.Command {
+	var opts create.CreateView
+	var immutableID int64
+
+	cmd := &cobra.Command{
+		Use:   "update [PROJECT_NAME]",
+		Short: "update immutable tag rule",
+		Long:  "update immutable tag rule for a project in harbor",
+		Args:  cobra.MaximumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var err error
+			var projectName string
+
+			if len(args) > 0 {
+				projectName = args[0]
+			} else {
+				projectName, err = prompt.GetProjectNameFromUser()
+				if err != nil {
+					return fmt.Errorf("failed to get project name: %v", utils.ParseHarborErrorMsg(err))
+				}
+			}
+
+			if immutableID == 0 {
+				immutableID = prompt.GetImmutableTagRule(projectName)
+			}
+
+			existingRule, err := getImmutableRuleByID(projectName, immutableID)
+			if err != nil {
+				return fmt.Errorf("failed to load existing immutable tag rule: %v", err)
+			}
+
+			updateView := buildUpdateViewFromRule(existingRule)
+			applyFlagOverrides(updateView, opts)
+
+			if !hasAnyUpdateFlag(opts) {
+				create.CreateImmutableView(updateView)
+			}
+
+			applyViewToRule(existingRule, updateView)
+
+			if existingRule.ID == 0 {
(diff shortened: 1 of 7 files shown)


cmd/harbor/root/tag/immutable/update.go
```diff
@@ -0,0 +1,169 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to update immutable tag rule: %v", err)
+			}
+
+			return nil
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.Int64VarP(&immutableID, "immutable-id", "", 0, "immutable rule ID to update")
```
NucleoFusion: Dont do ID, do tag_name
And then find the ID from that.

pkg/api/immutable_handler.go
```diff
@@ -77,3 +79,25 @@ func DeleteImmutable(projectName string, ImmutableID int64) error {
...
+	}
+	if rule == nil {
+		return errors.New("immutable rule payload cannot be nil")
+	}
+
+	_, err = client.Immutable.UpdateImmuRule(ctx, &immutable.UpdateImmuRuleParams{
+		ProjectNameOrID: projectName,
+		ImmutableRuleID: immutableID,
+		ImmutableRule:   rule,
+	})
```
NucleoFusion, line 101: Image

Use this so that the value is only represented as a name

NucleoFusion (changes requested):

cmd/harbor/root/tag/immutable/update.go
NishchayRajput: done!

pkg/api/immutable_handler.go
NishchayRajput: done~

qcserestipy:
Please refresh docs such that lint does not fail

NishchayRajput:
> Please refresh docs such that lint does not fail

done

bupd (changes requested):
please rebase

cmd/harbor/root/tag/immutable/update.go
```diff
@@ -0,0 +1,195 @@
...
+
+func UpdateImmutableCommand() *cobra.Command {
+	var opts create.CreateView
+	var tagName string
+
+	cmd := &cobra.Command{
+		Use:   "update [PROJECT_NAME]",
+		Short: "update immutable tag rule",
+		Long:  "update immutable tag rule for a project in harbor",
+		Args:  cobra.MaximumNArgs(1),
```
qcserestipy, line 36: add an example docs here please

cmd/harbor/root/tag/immutable/update.go
qcserestipy: maybe the flag should be called immutable-tag or so, at least please update the description that it has to be the name

qcserestipy (changes requested):
@NishchayRajput Thank you for taking this on, the interactive-first flow fits the rest of the CLI nicely and the happy path works as advertised.

Testing beyond the happy path surfaced a few issues that need to be addressed before merge:

1. **Partial updates are lossy**: `buildUpdateViewFromRule` only reads `TagSelectors[0]` and the first repository scope selector, and `applyViewToRule` rebuilds the PUT payload with exactly one selector each. A rule created via the API with multiple selectors silently loses all but the first on any update. Fix: preserve the selectors the user did not touch.

2. **Rule lookup by `--tag_name` is ambiguous.** Multiple rules can share a tag pattern (differing by repo scope), and the first match wins with no error. Please add an `--id` flag like the delete command uses so updates are deterministic. Also, the flag takes a pattern rather than a name and is the only snake_case flag on the command, so please rename it to `--tag-pattern` (which your PR description already uses).

3. **Cancel and error handling**: cancelling the rule picker, an empty project, or an unreachable Harbor all fall through to rule ID 0 and report "immutable rule not found" instead of the real cause, since `prompt.GetImmutableTagRule` swallows the `ListImmutable` error. Aborting the edit form additionally hits `log.Fatal` inside the shared create view, which kills the process and bypasses `RunE` error handling.

4. **`XIsResourceName` should be `true`**: the immutable handlers in `pkg/api/immutable_handler.go` hardcode it to `false` although a project name is always passed. Harbor allows purely numeric project names (e.g. `2024`), and with the header set to `false` the server parses those as a project ID and resolves a different project. To be fair, `false` matches the wire default, so this predates your PR, but since you are touching all four handlers anyway please send `true`, same as `webhook_handler.go` and `project_config_handler.go` already do.

Follow-up material, fine to defer:

- A flags-only update of a rule without repository scope selectors sends an empty `{Decoration:"", Pattern:""}` selector; worth a guard.
- The interactive path lists rules twice (picker, then `getImmutableRuleByID`); returning the selected rule directly saves a round-trip.
- `applyFlagOverrides` uses `!= ""` checks; the `flags.Changed()` pattern from `labels/update.go` would also allow clearing a value intentionally.

For verification: a rule with two tag selectors created via the API, a numeric project name, and ctrl+c in both the picker and the form cover the cases above.

Please also rebase onto main, the branch currently has conflicts. Happy to re-review after.

PR #810, opened 2026-04-15
fix: add more stable validation for harbor-cli login

### LINKED ISSUE #808: login fails with unhelpful swagger error when /users/current is unavailable (credentials are valid)
### Summary

`harbor login` validates credentials exclusively by calling `GET /api/v2.0/users/current` (via `client.User.GetCurrentUserInfo` in [`cmd/harbor/root/login.go`](https://github.com/goharbor/harbor-cli/blob/main/cmd/harbor/root/login.go#L136)). If that single endpoint is unavailable — even though the credentials are valid and the rest of the Harbor API is working — login fails with a cryptic go-swagger error that gives the user no useful signal.

### Repro

Against a Harbor instance where `/api/v2.0/users/current` happens to return 502 (the other endpoints work fine, basic auth with an OIDC CLI secret is accepted):

```
$ harbor login <server> -u <user> -p <cli-secret>
Error: login failed: response status code does not match any response statuses defined for this endpoint in the swagger spec (status 502): {}
```

Direct curl against the same server with the same credentials:

| Endpoint | Status |
|---|---|
| `GET /api/v2.0/systeminfo` | 200 — `auth_mode: oidc_auth`, `primary_auth_mode: true` |
| `GET /api/v2.0/projects` (basic auth) | **200** — credentials are valid |
| `GET /api/v2.0/users/current` (basic auth) | **502** |
| `GET /api/v2.0/users/current` (wrong password) | **502** |
| `GET /api/v2.0/users`, `/users/search` | **502** |

The 502 on `/users/*` is a server-side defect on that specific instance, but it exposes two problems in the CLI:

### Problems

1. **Fragile validation surface.** `harbor login` depends on a single, non-essential endpoint to decide whether credentials work. A user who can successfully `harbor project list`, `harbor repo list`, push/pull via Docker, etc. still cannot log in — because this one endpoint happens to be broken. `/users/current` is not needed for the login flow itself; it is used only as a \"ping with auth\" probe.

2. **Unhelpful error message.** The user sees the raw go-client/swagger string:
   ```
   response status code does not match any response statuses defined for this endpoint in the swagger spec (status 502): {}
   ```
   There is no indication that:
   - the HTTP status was 502 (server error, not auth error),
   - the error came from `/users/current` specifically,
   - the credentials themselves may still be valid.
   A user naturally concludes \"my password is wrong\" and keeps retrying — the actual cause is not discoverable without reading the CLI source.

### Suggested fixes

Any of these would address the issue; (1) and (3) are the most impactful:

1. **Use a more reliable probe for login validation** — e.g. `GET /api/v2.0/ping` (unauthenticated reachability) combined with an authenticated call that does not require the `/users` subtree to be healthy. `GET /api/v2.0/projects?page_size=1` would prove the credentials are accepted without depending on `/users/current`. Better: try multiple endpoints and accept success on any one of them.

2. **Distinguish transport/server errors from auth errors.** If the HTTP status is 5xx, surface it as \"server error (HTTP 5xx) contacting <endpoint>\" rather than a generic \"login failed\". A 401/403 means the credentials are wrong; a 502/503 means the server is sick — these deserve different messages and different exit codes.

3. **Improve `utils.ParseHarborErrorMsg`** to extract and display the HTTP status code and the endpoint that was called, instead of passing the raw go-swagger \"does not match any response statuses defined in the swagger spec\" text through to the user.

4. (Optional) add a `--skip-verify` flag to store the credentials without the pre-flight check, so a user who knows the server has a partial outage can still configure the CLI.

### Environment

- harbor-cli: `main` @ `da44eac` (built locally from source)
- Harbor server: `auth_mode=oidc_auth`, `primary_auth_mode=true`
- OS: macOS (darwin/arm64)

qcserestipy:
## Description

Adds a `--skip-verify-client` flag to `harbor login` and improves error handling during client validation.

- Fixes #808

## Type of Change

- [x] New feature

## Changes

- Added `--skip-verify-client` flag to skip credential validation against the Harbor server during login
- On 4xx errors, return a clear "authentication failed" message
- On 5xx errors, check `ListProjects` and `Ping` endpoints and report which succeeded/failed


cmd/harbor/root/login.go
```diff
@@ -200,3 +205,54 @@ func RunLogin(opts login.LoginView) error {
...
+	// 401/403 = definite auth failure
+	if errorCode == "401" || errorCode == "403" {
+		return fmt.Errorf("authentication failed, check your credentials: %v", utils.ParseHarborErrorMsg(err))
+	}
+
+	// For other errors (e.g. 412 for robot/OIDC accounts, 5xx),
+	// fall back to secondary endpoints to verify creds and reachability.
+	_, projectErr := client.Project.ListProjects(ctx, &project.ListProjectsParams{
+		Page:     func(v int64) *int64 { return &v }(int64(1)),
+		PageSize: func(v int64) *int64 { return &v }(int64(1)),
```
NucleoFusion: If the 1.26.0 bump goes through, we could use the `new` keyword here.
Just a comment to keep in mind, not a change.

EDIT: 
```suggestion
		Page:     new(int64(1)),
		PageSize: new(int64(1)),
```

The go version has been updated to 1.26
(if you would check it, it would be nice)

NucleoFusion (approved):
lgtm!

bupd (approved):
/lgtm

PR #817, opened 2026-04-18
chore: bump Go version to 1.26.2

### LINKED ISSUE #805: deps: bump module version to v1.26.2
## Overview

Currently the `vulnerability check` job is failing due to the incompatibility with `govulncheck` tool, *as the current version support go >= 1.25.0, but the current project module's go version is 1.24.8 ([ref](https://github.com/goharbor/harbor-cli/blob/main/go.mod#L3))*

The problem is noticed in this action ([ref](https://github.com/goharbor/harbor-cli/actions/runs/24400187260/job/71268372780#step:4:639)) where the current job pulls the latest version which is incompatible and causing the above error ([ref](https://github.com/goharbor/harbor-cli/blob/main/.dagger/vuln.go#L18)).

## Expected Solution

Need to bump the version from `1.24.8` to `1.26.2`
```diff
- go 1.24.8
+ go 1.26.2
```
> **NOTE:** Also, with compatibility considerations, just apart from updating the module, it is must to ensure that the project isn't broken, **as a major version bump may introduce breaking changes too.**

adityashekhawat1999:
## Description
Bumps Go version from 1.24.8 to 1.25.0

- Fixes #805 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update
- [x] Chore / maintenance

## Changes
- Updated Go version in `go.mod` to 1.25.0
- Ran `go mod tidy` to ensure dependency compatibility
 
## Note
While testing locally running `go mod vendor` introduced large changes in the `vendor/` directory which are not required for this PR as discussed in the issue, changes are limited to `go.mod` and `go.sum`.

Build and tests were verified without including vendor updates.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/goharbor/harbor-cli
 
-go 1.25.0
+go 1.26.2
 
 require (
 	github.com/atotto/clipboard v0.1.4
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/goharbor/harbor-cli
 
-go 1.24.8
+go 1.25.0
 
 require (
 	github.com/atotto/clipboard v0.1.4


NucleoFusion:
Not a review, just a question.
Why arent we going 1.26 instead of only 1.25?

adityashekhawat1999:
> Not a review, just a question. Why arent we going 1.26 instead of only 1.25?

I was also thinking about that but none of the maintainers mentioned it in the issue, i think we should discuss it in the next meet

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/goharbor/harbor-cli
 
-go 1.24.8
+go 1.25.0
```
bupd:
```suggestion
go 1.26.2
```

bupd (changes requested):
lets jump to 1.26.2

PR #820, opened 2026-04-19
fix(user): resolve missing id parsing and pagination in user executio…

### LINKED ISSUE #819: [bug]: Fix user lookup bug causing operations on UserID 0 and ignoring pagination
## Description
There is a critical issue in the `harbor user delete`, `elevate`, and `password` commands where user lookup by name is unreliable and may result in unintended operations on `UserID 0`.

The CLI claims to support "name or id", but:
- User lookup only checks the first page of results (pagination is ignored)
- If a user is not found, the function returns `(0, nil)` instead of an error
- This causes destructive operations (e.g., delete) to execute on `UserID 0`
- Numeric inputs are not treated as IDs, but as usernames

This leads to incorrect behavior and potential safety risks.

## Steps to Reproduce
Case 1: Non-existent username
1. Run command:
   harbor user delete some_nonexistent_user
2. Observe behavior

Case 2: User exists but not on first page
1. Ensure the target user is not present on page 1 of ListUsers
2. Run command:
   harbor user delete <valid_username>
3. Observe behavior

Case 3: Using numeric ID
1. Run command:
   harbor user delete 123
2. Observe behavior

## Expected Behavior
- The CLI should search across all pages or use server-side filtering
- It should return an error if the user is not found
- Numeric input should be interpreted as a user ID
- No operation should be executed when the user cannot be resolved

## Actual Behavior
- Only the first page of users is searched
- If the user is not found, the function returns UserID 0 with no error
- Commands proceed with destructive API calls using UserID 0
- Numeric inputs are treated as usernames instead of IDs

## Environment
- OS: (e.g., Ubuntu 22.04 / Windows 11)
- Tool version: (e.g., harbor CLI vX.X.X)
- Other relevant details: N/A

## Additional Context
Root cause:
- api.GetUsersIdByName does not handle pagination
- It returns (0, nil) when no match is found
- CLI commands do not validate the returned ID before making API calls
- No parsing is done to distinguish numeric IDs from usernames

Suggested fixes:
- Implement pagination in user lookup
- Return explicit error when user is not found
- Parse numeric arguments as IDs before falling back to name lookup

PrasunaEnumarthy:
## Description
This pull request completely resolves the critical ID resolution and execution bugs within the user management commands (`delete`, `elevate`, `password`). 
Previously, `api.GetUsersIdByName` silently ignored pagination (locking validation to page 1) and silently returned `UserID 0` on failures instead of throwing an error. This caused valid missing-user queries to inadvertently execute destructive operations on UserID 0. Furthermore, the CLI commands explicitly advertised support for numeric IDs, but failed to actually parse them, causing ID lookups to fail and default to mutating UserID 0. 
This PR properly paginates the background lookup, cleanly errors on unresolvable entries, and implements strict `strconv.ParseInt` checks.

- Fixes #819 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
- **Implemented API Pagination:** `pkg/api/user_handler.go:GetUsersIdByName` now iterates through subsequent pages up to completion to fully ensure the username being queried exists on the system.
- **Fail-Safe Integrity:** `api.GetUsersIdByName` now correctly throws `fmt.Errorf` on unresolved query strings instead of returning a completely silent `0, nil`.
- **Enforced CLI Argument Contract:** Updated `cmd/harbor/root/user`'s `delete.go`, `elevate.go`, and `password.go`. Commands now explicitly check if the argument is a parsable integer ID before aggressively falling back to the name-search index.


cmd/harbor/root/user/elevate.go
```diff
@@ -35,16 +36,20 @@ func ElevateUser(args []string) error {
...
-			return err
-		}
-		if userId == 0 {
-			err = fmt.Errorf("User with name '%s' not found", args[0])
-			log.Error(err.Error())
-			return err
+		if parsedID, parseErr := strconv.ParseInt(args[0], 10, 64); parseErr == nil {
+			userId = parsedID
+		} else {
+			userId, err = getUsersIDByName(args[0])
```
NucleoFusion: This is prone to not giving the right output since harbor allows names like `111` or `1234`.
hence its much better to add a `--id` flag to signal that the provided `args[0]` is an id. Else by default, we do names.

cmd/harbor/root/user/elevate.go
```diff
@@ -35,16 +36,20 @@ func ElevateUser(args []string) error {
...
+			if err != nil {
+				err = fmt.Errorf("failed to get user id for '%s': %v", args[0], err)
+				log.Error(err.Error())
+				return err
+			}
+			if userId == 0 {
+				err = fmt.Errorf("user '%s' not found", args[0])
+				log.Error(err.Error())
```
NucleoFusion: You can just return errors. Cobra logs them at the root.

cmd/harbor/root/user/password.go
```diff
@@ -42,12 +43,16 @@ func UserPasswordChangeCmd() *cobra.Command {
...
-				}
-				if userId == 0 {
-					return fmt.Errorf("User with name '%s' not found", args[0])
+				// Try parsing ID first
+				if parsedID, parseErr := strconv.ParseInt(args[0], 10, 64); parseErr == nil {
+					userId = parsedID
+				} else {
+					userId, err = api.GetUsersIdByName(args[0])
+					if err != nil {
+						err = fmt.Errorf("failed to get user id for '%s': %v", args[0], err)
```
NucleoFusion: This is prone to not giving the right output since harbor allows names like 111 or 1234.
hence its much better to add a --id flag to signal that the provided args[0] is an id. Else by default, we do names.

pkg/api/user_handler.go
```diff
@@ -106,18 +107,30 @@ func ListUsers(opts ...ListFlags) (*user.ListUsersOK, error) {
...
+		for _, user := range u.Payload {
+			if user.Username == userName {
+				return user.UserID, nil
+			}
+		}
+		// Stop requesting if we received less items than we asked for
+		if int64(len(u.Payload)) < pageSize {
+			break
+		}
+		page++
```
NucleoFusion, line 132: We can simplify this a lot by just using the `q` param.

Image

You can reuse the `utils.BuildQueryParam` to create a custom one for you, if you dont want to hardcode (prolly shouldnt).

NucleoFusion (changes requested):

PrasunaEnumarthy:
@NucleoFusion I’ve addressed the suggestions by removing ambiguity with a `--id` flag and updating all related commands,please let me know if anything else needs to be fixed.

cmd/harbor/root/user/elevate.go
```diff
@@ -31,49 +31,47 @@ var (
...
+		return errors.New("User did not confirm elevation. Aborting command.")
 	}
 
 	err = elevateUserAPI(userId)
 	if err != nil {
 		if isUnauthorizedError(err) {
-			err = errors.New("Permission denied: Admin privileges are required to execute this command.")
-		} else {
-			err = fmt.Errorf("failed to elevate user: %v", err)
+			return errors.New("Permission denied: Admin privileges are required to execute this command.")
```
NucleoFusion: you can use the fmt.Errorf
but return it instead of print.

The current change with `errors.New` loses the error message itself.

NucleoFusion (changes requested):

NucleoFusion (changes requested):
@PrasunaEnumarthy lint seems to be failing, most likely is just a rebasing issue. Just rebase this so its up to date.
Also, on a side not, consider marking review comments as resolved, that makes it easy to navigate the PR and understand whats solved and whats not.

PrasunaEnumarthy:
> @PrasunaEnumarthy lint seems to be failing, most likely is just a rebasing issue. Just rebase this so its up to date.
> Also, on a side not, consider marking review comments as resolved, that makes it easy to navigate the PR and understand whats solved and whats not.

Yeah , i will rebase the branch and resolve the addressed review comments

bupd (changes requested):
@PrasunaEnumarthy please rebase and resolve conflicts

PrasunaEnumarthy:
@bupd i have resolved all the comments and rebased the branch... can you ptal whenever you have time, thanks

bupd:
stale

PR #841, opened 2026-04-24
fix(registry): handle missing registry lookup in delete by name

### LINKED ISSUE #840: [bug]: registry lookup by name can silently fail and delete may use ID 0
## Description
`GetRegistryIdByName()` can fail silently in two ways:

It returns `(0, nil)` when the registry name is not found.
It calls `ListRegistries()` without a query filter, so lookup depends on paginated results and may miss registries beyond the first page.

This becomes user-visible in `harbor registry delete <registry-name>` because the command ignores lookup errors and may attempt to delete registry ID 0.

## Steps to Reproduce
1. Ensure multiple registries exist (enough to trigger pagination)
2. Run command:
harbor registry delete <nonexistent-name>
3. Observe error
- Lookup may silently fail
- Delete may proceed with registry ID 0

## Expected Behavior
- Lookup should use exact-name filtering via query
- Missing registry should return:
   registry '<name>' not found
- Delete command should stop immediately if lookup fails

## Actual Behavior
- Lookup returns (0, nil) when registry is not found
- Pagination may cause valid registries to be missed
- Delete command may proceed using ID `0`

## Environment
- OS:  (e.g., Linux / macOS / Windows)
- Tool version: (e.g., Harbor CLI version)
- Other relevant details: N/A

## Additional Context
Affected files:
- `pkg/api/registry_handler.go`
- `cmd/harbor/root/registry/delete.go`
Similar issue fixed for users in PR #820

Image

PrasunaEnumarthy:
## Description
This PR fixes registry name lookup so it no longer fails silently when a registry is missing or falls beyond the first page of results.

It also updates `harbor registry delete <name>` to stop immediately when name resolution fails instead of proceeding with registry ID `0`.

- Fixes #840 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
- query registries by exact name in GetRegistryIdByName
- return registry '<name>' not found when lookup fails
- handle lookup errors in registry delete
- add focused unit tests for lookup and delete error handling


cmd/harbor/root/registry/delete.go
```diff
@@ -22,6 +22,11 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	getRegistryIDByNameFunc = api.GetRegistryIdByName
+	deleteRegistryFunc      = api.DeleteRegistry
+)
+
```
NucleoFusion: This testing strategy is wrong, please refer to this,
#797 While it isnt merged, you can decouple the TestCmd function to find out how to test

NucleoFusion (changes requested):

qcserestipy (changes requested):
This PR fixes issue, but also introduces new architectural patterns not used elsewhere in this project:

- `newDeleteRegistryCommand(deps ...)` constructor pattern
- `deleteRegistryDeps` DI struct in command layer
- mutable package-level function seam (`listRegistriesFunc`) in API layer

I think this goes beyond scope of #840 and breaks local consistency (other commands do not use this style).

Can we keep this fix minimal and follow existing patterns?
- keep `DeleteRegistryCommand()` style consistent with other commands
- handle `GetRegistryIdByName` error directly in current flow
- add focused tests without introducing a new command-constructor/DI pattern

PrasunaEnumarthy:
> This PR fixes issue, but also introduces new architectural patterns not used elsewhere in this project:
> 
> * `newDeleteRegistryCommand(deps ...)` constructor pattern
> * `deleteRegistryDeps` DI struct in command layer
> * mutable package-level function seam (`listRegistriesFunc`) in API layer
> 
> I think this goes beyond scope of #840 and breaks local consistency (other commands do not use this style).
> 
> Can we keep this fix minimal and follow existing patterns?
> 
> * keep `DeleteRegistryCommand()` style consistent with other commands
> * handle `GetRegistryIdByName` error directly in current flow
> * add focused tests without introducing a new command-constructor/DI pattern

Hi @qcserestipy , I’ve made the requested changes... removed the DI/constructor setup, cleaned up the seam, aligned the command with existing patterns, and simplified the tests.

The lint issues seem to be pre-existing (from a golangci-lint update) and aren’t related to this PR. Happy to fix them in a separate PR if you’d prefer.

pkg/api/registry_handler.go
```diff
@@ -175,18 +175,31 @@ func GetRegistryProviders() ([]string, error) {
 }
```
qcserestipy, line 175: abstracting this function away is not necessary

cmd/harbor/root/registry/delete_test.go
```diff
@@ -0,0 +1,28 @@
...
+package registry
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDeleteRegistryCommand_Metadata(t *testing.T) {
+	cmd := DeleteRegistryCommand()
```
qcserestipy, line 23: Please adapat this test to only test for correctness of input parameters and expected outputs. E2E tests will be enabled after a local harbor instance is available

qcserestipy (changes requested):
Thank you for your contribution, please adjust the code according to the comments. Please also do not change the layout of the actual command. Do now create unneeded types and do not add uneeded constructor functions. Please stick to consistent format of cobra commands as for others.

PrasunaEnumarthy:
@qcserestipy , thanks for the review! Sorry for the late response. I've addressed all the feedback:
- Removed abstractions: `GetRegistryIdByName` now uses `ListFlags{Name: registryName}` directly, matching existing lookup patterns.
- No layout changes: `DeleteRegistryCommand()` follows the standard project Cobra pattern with no DI, constructors, or new types.
- Updated tests: Simplified `TestDeleteRegistryCommand` to focus strictly on metadata and parameters, removing all E2E behavior.

bupd (approved):
/lgtm

PR #843, opened 2026-04-28
perf: fix regex compilation, string building, and concurrency hotpaths

bupd:
## Summary
- **Pre-compile regexes**: Extract 12 `regexp.Compile`/`regexp.MatchString` calls in `pkg/utils/` to package-level `regexp.MustCompile` variables, eliminating repeated compilation (0 allocs/op in benchmarks)
- **Add API timeout**: Add 30s default timeout to `ContextWithClient()` to prevent indefinite API call hangs
- **Fix O(n^2) string building**: Replace `s +=` concatenation with `strings.Builder` in multiselect `listView()` and build hyperlinks during row construction in vulnerability list instead of post-processing with `strings.ReplaceAll`
- **Fix N+1 API query**: Batch-resolve project IDs to names in vulnerability summary `renderDangerousArtifacts()` instead of calling `GetProject` per artifact
- **Pre-allocate slices**: Add capacity hints where length is known (vulnerability list, artifact list, member list, robot permissions)
- **Bounded concurrency**: Add semaphore (max 5 concurrent) to project delete to prevent unbounded goroutine spawning
- **Pagination guards**: Add max iteration limits (1000 pages) to pagination loops in vulnerability and quota handlers

## Test plan
- [x] All existing tests pass (`go test ./...`)
- [x] Benchmark tests added for regex validation hot paths — confirm 0 allocs/op
- [ ] Manual test: `harbor vulnerability list` still renders CVE hyperlinks correctly
- [ ] Manual test: `harbor project delete proj1 proj2 ... projN` with N > 5 projects

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
*Automated by [nightshift](https://github.com/marcus/nightshift)*

### DIFF
--- a/pkg/views/vulnerability/list/view.go
+++ b/pkg/views/vulnerability/list/view.go
@@ -44,17 +44,15 @@ func ViewVulnerabilityList(vulnerabilities []*models.VulnerabilityItem, hasNext 
 		return
 	}
 
-	var rows []table.Row
-	cveLinks := make(map[string]string)
+	rows := make([]table.Row, 0, len(vulnerabilities))
 	for _, vulnerability := range vulnerabilities {
-		link := ""
+		cveDisplay := vulnerability.CVEID
 		if len(vulnerability.Links) > 0 {
-			link = vulnerability.Links[0]
+			cveDisplay = ansi.SetHyperlink(vulnerability.Links[0]) + vulnerability.CVEID + ansi.ResetHyperlink()
 		}
 
-		cveLinks[vulnerability.CVEID] = link
 		rows = append(rows, table.Row{
-			vulnerability.CVEID,
+			cveDisplay,
 			vulnerability.Desc,
 			vulnerability.RepositoryName,
 			vulnerability.Digest,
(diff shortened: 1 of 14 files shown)


pkg/views/vulnerability/list/view.go
NucleoFusion, line 1: Something is breaking output here,

Image

NucleoFusion (changes requested):

pkg/views/vulnerability/list/view.go
NucleoFusion: It seems to happen because the model runs a `.Width()` on the content and includes the ansi sequeances in calc too.
And then it tries to shorten it leading to half a link and break

NucleoFusion: The previous code is the best possible way I beleive.
You add the links in print-time rather than before model creation.
Reverting changes to this file would solve it.

NucleoFusion (changes requested):

bupd:
closing this as it introduces more issues than it actually solves.

bupd:
thanks @NucleoFusion for reviewing it.

PR #849, opened 2026-05-01
fix(login): enforce mutual exclusivity of --password and --password-stdin flags

### LINKED ISSUE #848: fix(login): enforce mutual exclusivity of --password and --password-stdin
## Description

The `harbor login` command currently allows both `--password` (`-p`) and `--password-stdin` to be used together, resulting in ambiguous behavior where one input silently overrides the other without any user-facing error.

This behavior may lead to confusion in CI/CD pipelines where unintended input sources are used.

When both flags are provided, `--password-stdin` takes precedence and discards the value passed via `--password`, leaving the user with no indication that their `-p` input was ignored. This can lead to confusing and unexpected authentication behavior.

Additionally, `--context-name` is the only login flag that lacks a short alias (`-n`), which is inconsistent with other flags such as `-u` and `-p`.

## Steps to Reproduce

1. Run command:
   ```
   harbor login demo.goharbor.io -u admin -p MyPassword123 --password-stdin
   ```
2. With arguments/config: any valid server and username
3. Observe: the command does **not** return an error — it silently ignores `-p` and waits for stdin input instead

## Expected Behavior

The CLI should immediately return a clear error when both `--password` and `--password-stdin` are provided together:

```
Error: if any flags in the group [password password-stdin] are set none of the others can be; [password password-stdin] were all set
```

This is consistent with the behavior of `docker login` and other standard CLI tools.

## Actual Behavior

No error is returned. The `--password-stdin` path takes precedence silently, discarding the value passed via `--password`, leading to confusing and unpredictable behavior.

## Environment

- OS: Linux / macOS / Windows (reproducible on all platforms)
- Tool version: latest (`main` branch)
- Other relevant details: Cobra v1.9+ supports `MarkFlagsMutuallyExclusive` natively

## Additional Context

**Proposed fix:**

Use Cobra's built-in `MarkFlagsMutuallyExclusive` after flag definitions in `LoginCommand()`:

```go
cmd.MarkFlagsMutuallyExclusive("password", "password-stdin")
```

This enforces the constraint at the framework level with zero additional logic, and also updates the `--help` output automatically to communicate the restriction to users.

Also propose adding `-n` as a short alias for `--context-name` to maintain flag consistency across the login command.

**Files affected:**
- `cmd/harbor/root/login.go`
- `cmd/harbor/root/login_test.go` (new test case)

gcharpe1604:
## Description

The `harbor login` command accepted both `--password` and `--password-stdin` simultaneously without any error. This caused silent ambiguous behavior where `--password-stdin` overwrote the `-p` value with no user-facing warning.

- Fixes #848

## Type of Change

- [x] Bug fix

## Changes

- Added `cmd.MarkFlagsMutuallyExclusive("password", "password-stdin")` in `LoginCommand()` — uses Cobra's built-in constraint enforcement
- Added `-n` as a short alias for `--context-name` to match convention of all other login flags
- Added test case `Test_Login_Failure_MutuallyExclusiveFlags` to cover the validation

## Testing

Verified locally:

### DIFF
--- a/cmd/harbor/root/login.go
+++ b/cmd/harbor/root/login.go
@@ -80,9 +80,11 @@ func LoginCommand() *cobra.Command {
 
 	flags := cmd.Flags()
 	flags.StringVarP(&Username, "username", "u", "", "Username")
-	flags.StringVarP(&Name, "context-name", "", "", "Login context name (optional)")
+	flags.StringVarP(&Name, "context-name", "n", "", "Login context name (optional)")
 	flags.StringVarP(&Password, "password", "p", "", "Password")
 	flags.BoolVar(&passwordStdin, "password-stdin", false, "Take the password from stdin")
+
+	cmd.MarkFlagsMutuallyExclusive("password", "password-stdin")
 
 	return cmd
 }
--- a/cmd/harbor/root/login_test.go
+++ b/cmd/harbor/root/login_test.go
@@ -106,3 +106,19 @@ func Test_Login_Success_RobotAccount(t *testing.T) {
 	err := cmd.Execute()
 	assert.NoError(t, err, "Expected no error for robot account login")
 }
+
+func Test_Login_Failure_MutuallyExclusiveFlags(t *testing.T) {
+	tempDir := t.TempDir()
+	data := helpers.Initialize(t, tempDir)
+	defer helpers.ConfigCleanup(t, data)
+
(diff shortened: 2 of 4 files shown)


NucleoFusion (approved):
lgtm!
Thanks for the contribution

qcserestipy (approved):

bupd (approved):
/lgtm

PR #816, opened 2026-04-17
fix(user): add email format validation to user create command

PrasunaEnumarthy:
## Description
Adds missing local input validation to the `harbor user create` command to prevent syntactically invalid email addresses from being blindly passed to the Harbor API. 
This ensures we "fail fast" and provide clean, immediate feedback on the CLI instead of relying on the backend to reject malformed data.

- Fixes #815 

## Type of Change
Please select the relevant type.

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
- Imported `net/mail` package to natively parse and validate email strings.
- Added a pre-execution validation step in `cmd/harbor/root/user/create.go` for the `--email` flag.
- Ensured the CLI immediately returns `error: invalid email format...` before initiating network requests to the API.

## Proof of Validation Catching Errors
image

PR #850, opened 2026-05-02
fix(user): validate email format before API call in create command

### LINKED ISSUE #815: [bug]: Missing email validation in `harbor user create` command
## Description
The harbor user create command accepts an --email flag but does not validate the input format before sending the request to the API.

Currently, invalid email values are accepted by the CLI and passed directly to the Harbor API, resulting in delayed API errors instead of early validation at the CLI level

## Steps to Reproduce
Run the following command with an invalid email:
```
harbor user create --username bob --password Secret123! --email "not_an_email" 
```
## Expected Behavior
The CLI should validate the email format locally and return a clear error message before making the API request.

Example:

Error: invalid email format

## Actual Behavior
The CLI accepts the invalid email and sends the request to the API, which may result in a less clear or delayed error response.

## Environment
- OS: Any
- Tool version: Latest (or current main branch)
- Other relevant details: Not environment-specific

## Additional Context
Adding early validation improves CLI usability by:

- Preventing unnecessary API calls
- Providing faster feedback to users
- Aligning with existing validation patterns in other commands

madmecodes:
Closes #815

When all flags are provided, `harbor user create` skips the interactive
form and calls the API directly. The interactive form validates email
via `utils.ValidateEmail()`, but the flag path didn't, so invalid
emails were sent to the API and returned confusing server-side errors.

Added the same `ValidateEmail` check to the flag path so the CLI
rejects bad emails early with a clear message.

### DIFF
--- a/cmd/harbor/root/user/create.go
+++ b/cmd/harbor/root/user/create.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 
 	"github.com/goharbor/harbor-cli/pkg/api"
+	"github.com/goharbor/harbor-cli/pkg/utils"
 
 	"github.com/goharbor/harbor-cli/pkg/views/user/create"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


qcserestipy:
This is a duplicate of already approved #816

madmecodes:
Closing - this duplicates #816 which is already approved.

PR #851, opened 2026-05-02
fix(registry): return error on failed name lookup and handle it in delete

### LINKED ISSUE #840: [bug]: registry lookup by name can silently fail and delete may use ID 0

madmecodes:
Closes #840

Two bugs in `GetRegistryIdByName`:

1. Returns `(0, nil)` when the registry name isn't found, so callers
   can't distinguish "not found" from success. Fixed to return
   `fmt.Errorf("registry '<name>' not found")`.

2. Calls `ListRegistries` with empty opts, so the lookup only checks
   the default page. Now passes the name filter so the API handles
   filtering server-side.

Also fixed `registry delete` to:
- Actually check the error from `GetRegistryIdByName` instead of
  discarding it with `_`
- Try `strconv.ParseInt` first so numeric IDs work as arguments too

Same pattern as the user lookup fix in #820.

### DIFF
--- a/cmd/harbor/root/registry/delete.go
+++ b/cmd/harbor/root/registry/delete.go
@@ -15,6 +15,7 @@ package registry
 
 import (
 	"fmt"
+	"strconv"
 	"sync"
 
 	"github.com/goharbor/harbor-cli/pkg/api"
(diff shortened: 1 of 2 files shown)


qcserestipy:
This is a duplicate of #841, this will be decided on later.

EDIT: Also please fix linting issues.

madmecodes:
Closing - duplicates #841.

Issue #821, opened 2026-04-20
[Tracker] Bubbletea Model Refactor (Loading + Init/Tick Integration)


NucleoFusion:
## Overview
Re-pickup and refactor of the Bubbletea model to better utilize built-in features like `tea.Init()` and `tea.Tick()` for improved loading states and UX consistency.

## Goals
- Introduce proper loading states using Bubbletea primitives
- Standardize async behavior across views
- Prepare for a unified table/list abstraction

## Tasks
- [ ] Utilize `tea.Init()` and `tea.Tick()` to implement loading states
- [ ] Identify and refactor commands/views that should display loading
- [ ] Convert existing components to use `loadingtablelist`
- [ ] Once migration is complete, rename `loadingtablelist` → `tablelist`

## Model Refactor Tracker
Track which models have been migrated to the new loading + init/tick pattern:

- [ ] tablelist
- [ ] selection
- [ ] multiselect
- [ ] tablegrid

> Mark each as complete once it fully adopts the new pattern (loading state + proper init/tick usage).

## Reference
- https://github.com/goharbor/harbor-cli/issues/756

## Notes
This is a tracker issue.

A more expansive issue will follow, covering the full scope of the problem and the complete refactor plan.  
This issue acts as a prequel — setting up the groundwork and introducing initial structural changes.

## Ownership
This is being actively worked on by @NucleoFusion

bupd:
Lets have this for LFX T2 26

NucleoFusion:
Sure,
I will then have a doc up for it

NucleoFusion:
@bupd What would the scope for that be?
Should I convert the other models for loading as well? or would that be for lfx?

digvijay-y:
@bupd  @NucleoFusion  I'm interested in working on this, I'll be applying through LFX, but is there anything I should take into consideration before applying. Also, I would like to get familiar, if there are any issues available to grab, happy to contribute. 

btw ,is @NucleoFusion  Already working on this?

NucleoFusion:
I am not, I am doing the basic prerequisite, and generating a doc for how to proceed with others as well for reference 

To get familiar, just go through the codebase once, understanding how it's structured. And then maybe learn a little about bubbletea and its v2 if not already familiar.

mundele2004:
Hi @bupd, @vad1mo, and @OrlinVasilev,

[...]

PR #852, opened 2026-05-02
refactor(views): add loading spinner to tablelist model

pie-314:
Implements async loading for the tablelist component. This allows the CLI to show a spinner immediately while data fetches in the background, improving UX.

   - Relates to #821

  Type of Change
   - [x] Refactor

  Changes
   - Added spinner.Model and IsLoading state to tablelist.
   - Updated Update() to handle async data and spinner animations.
   - Updated View() to show the spinner during loading.

### DIFF
--- a/pkg/views/base/tablelist/model.go
+++ b/pkg/views/base/tablelist/model.go
@@ -14,6 +14,7 @@
 package tablelist
 
 import (
+	"github.com/charmbracelet/bubbles/spinner"
 	"github.com/charmbracelet/bubbles/table"
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/lipgloss"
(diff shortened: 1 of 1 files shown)


NucleoFusion:
This does nothing.
What rows provides is always populated, and if rows actually has length 0. Then that just means there were actually no rows, not that its loading. What is `case []table.Row`? if you never update it, via any message. It will remain the same it does at declaration. So if rows _actually_ is 0 - which it shouldnt be 99% of the time - it _stays_ 0. This breaks the model.
Not to mention, this is a duplicate, we have a PR that adresses this issue already. 

@bupd @qcserestipy can we get this closed?

pie-314:
I understand the detailed investigation, @NucleoFusion. It’s clear to me that the synchronous handler pattern prevents us from refactoring at the moment. Closing this issue to maintain the cleanliness of the repository and continue our work on safer deletions in #856.

PR #853, opened 2026-05-02
fix(robot): exit cleanly when user presses q in permission grid

### LINKED ISSUE #847: [bug]: quit does not exit command in `robot create`
## Description

When in the multiselect page, where we select permissions, if we cancel using the `q` character, the program still executes and takes what we might have selected.
If nothing is selected, it shows `no permissions selected`, if we select something and then quit, it goes on ahead with no issues.

## Steps to Reproduce
1. Run command: `harbor-cli robot create`
2. With arguments/config: none
3. Fill placeholder name and description, and proceed to permission selection page
4. Select a permission but instead of moving forward, press `q` to quit.
5. The program continues and doesnt exist

## What you expected to happen.

Expected program to exit, and show a `user exited at system permission` message or similar.

## Actual Behavior

Continues to creation. Creates it successfully if you go through and fill rest of details.

## Environment
- OS: Fedora Linux
- Tool version: https://github.com/goharbor/harbor-cli/commit/455293c23d28c5fc6e67cb3bff7107d6cdf8403a

Famous077:
#Description:
This PR fixes a bug in the interactive permission selection step of harbor robot create and harbor project robot create.
When a user pressed q or Ctrl+C to back out of the permission grid, the TUI would close as expected : but the cancellation was silently swallowed. The calling code in ListPermissions had no way to know the user had opted out, so it would carry on and attempt to create the robot account anyway, sometimes with empty or incomplete permissions. Not great.
The fix is straightforward: a Cancelled field is added to the TableGrid struct, set to true when the user exits early, and checked in ListPermissions before any permission data is processed. If cancelled, an error is sent back up the chain and the command exits cleanly : no robot account created, no silent failures.

Fixes #847

Type of Change
 [x]Bug fix

#Changes:
Added a Cancelled bool field to the TableGrid struct in pkg/views/base/tablegrid/model.go
Set m.Cancelled = true in the Update() method when the user presses q or Ctrl+C, prior to calling tea.Quit
Added a grid.Cancelled check in ListPermissions (pkg/views/robot/select/view.go) : cancellation now sends an error on the result channel, halting robot creation cleanly

#Testing 
Run harbor robot create
Fill in name and duration
When the permission grid appears, press q
Command exits cleanly with: user exited at permission selection
Confirmed no robot account is created on the Harbor server

### DIFF
--- a/pkg/views/robot/select/view.go
+++ b/pkg/views/robot/select/view.go
@@ -132,31 +132,18 @@ func ListPermissions(perms *models.Permissions, kind string, ch chan<- Permissio
 	}
 	_, err = tea.NewProgram(grid, tea.WithAltScreen()).Run()
 	if err != nil {
-    fmt.Println("error creating permissions grid:", err)
-    ch <- PermissionSelectResult{
-        Permissions: nil,
-        Err:         err,
-    }
-    return
-}
-	if grid.Cancelled {
-    ch <- PermissionSelectResult{
-        Permissions: nil,
-        Err:         fmt.Errorf("user exited at permission selection"),
-    }
-    return
+		ch <- PermissionSelectResult{Permissions: nil, Err: err}
+		return
 	}
-	if err != nil {
-		fmt.Println("error creating permissions grid:", err)
+	if grid.Cancelled {
 		ch <- PermissionSelectResult{
 			Permissions: nil,
-			Err:         err,
+			Err:         fmt.Errorf("user exited at permission selection"),
 		}
 		return
 	}
-	data:= grid.GetData()
+	data := grid.GetData()
 	selectedPerms := []models.Permission{}
-
 	for rowIdx, displayName := range grid.RowLabels {
 		kebabResource := utils.ToKebabCase(displayName)
 
--- a/pkg/views/robot/select/view.go
+++ b/pkg/views/robot/select/view.go
@@ -1,4 +1,4 @@
-// Copyright Project Harbor Authors
+﻿// Copyright Project Harbor Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -145,7 +145,8 @@ func ListPermissions(perms *models.Permissions, kind string, ch chan<- Permissio
         Err:         fmt.Errorf("user exited at permission selection"),
     }
     return
-}
+	}
+	data:= grid.GetData()
 	if err != nil {
 		fmt.Println("error creating permissions grid:", err)
 		ch <- PermissionSelectResult{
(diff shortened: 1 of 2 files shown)


Famous077:
Hi @NucleoFusion , Can you review this PR whenever you get time.

NucleoFusion (changes requested):
There is a lot of whitespace changes, possibly from conflict on whether tabs are spaces or not.
Anyways, this makes it really hard for me to determine what actually changed, I cant really review this.

NucleoFusion:
And remove this from draft if this is completed. We dont look at draft PRs, only merge-able ones.

Famous077:
Hi @NucleoFusion. Now, I have added clear description of what i did here to resolve this issue, and along with that I mistakely forgot to remove the whitespaces which I corrected. Can you review whenever you get time. Love to take feedback if anything else needs to be implemented.

pkg/views/robot/select/view.go
```diff
@@ -131,6 +131,13 @@ func ListPermissions(perms *models.Permissions, kind string, ch chan<- Permissio
 		return
 	}
 	_, err = tea.NewProgram(grid, tea.WithAltScreen()).Run()
+	if grid.Cancelled {
+		ch <- PermissionSelectResult{
+			Permissions: nil,
+			Err:         fmt.Errorf("user exited at permission selection"),
+		}
+		return
+	}
```
NucleoFusion: Error should be checked first, and then `cancelled`
You can just re-order this. Put this if statement after the error check.

NucleoFusion (changes requested):

pkg/views/robot/select/view.go
```diff
@@ -131,6 +131,21 @@ func ListPermissions(perms *models.Permissions, kind string, ch chan<- Permissio
...
+    }
+    return
+}
+	if grid.Cancelled {
+    ch <- PermissionSelectResult{
+        Permissions: nil,
+        Err:         fmt.Errorf("user exited at permission selection"),
+    }
+    return
+}
```
NucleoFusion: Why is this added here? This breaks I assume

NucleoFusion (changes requested):

NucleoFusion (changes requested):
This PR is only supposed to change / work on the permission grid.
What are all these changes?

NucleoFusion:
The PR is as simple as adding a,
```go
	if grid.Cancelled {
    ch <- PermissionSelectResult{
        Permissions: nil,
        Err:         fmt.Errorf("user exited at permission selection"),
    }
    return
    ```
    In robot/select/view.go. After the model run
    
    And adding the relevant `Cancelled` field in the model itself, and then setting it as `Cancelled = true` in the "q" case in Update()

Famous077:
Now, I have implemented according to your suggestions. Can you please review and let me know if anything else needs to be implemented @NucleoFusion .

NucleoFusion (approved):
lgtm!
Thanks for the contribution!

PR #857, opened 2026-05-03
fix(password): handle credential errors and add flag support

### LINKED ISSUE #707: bug: user password command not working
Due passing of empty `PasswordChangeView`, password change does not happen via the CLI 
```
func UserPasswordChangeCmd() *cobra.Command {
	var opts reset.PasswordChangeView

	cmd := &cobra.Command{
		Use:   "password",
		Short: "Reset user password by name or id",
		Long:  "Allows admin to reset the password for a specified user or select interactively if no username is provided.",
		Args:  cobra.MinimumNArgs(0),
			var userId int64
			var err error
			resetView := &reset.PasswordChangeView{
				NewPassword:     opts.NewPassword,
				ConfirmPassword: opts.ConfirmPassword,
			}

			if len(args) > 0 {
				userId, err = api.GetUsersIdByName(args[0])
				if err != nil {
					log.Errorf("failed to get user id for '%s': %v", args[0], err)
					return
				}
				if userId == 0 {
					log.Errorf("User with name '%s' not found", args[0])
					return
				}
			} else {
				userId = prompt.GetUserIdFromUser()
			}

			reset.ChangePasswordView(resetView)

			err = api.ResetPassword(userId, opts)
			if err != nil {
				if isUnauthorizedError(err) {
					log.Error("Permission denied: Admin privileges are required to execute this command.")
				} else {
					log.Errorf("failed to reset user password: %v", err)
				}
			}
		},
	}
	return cmd
}
```
Note the empty `opts` getting passed to the API (`err = api.ResetPassword(userId, opts)`)

### LINKED ISSUE #742: [bug]: password command: ignored GetCredentials error leads to fatal exit
## Description
When running the password command, the code ignores the error returned by `GetCredentials` and proceeds with a zero-value credential. `UpdateCredentialsInConfigFile` then calls `log.Fatalf(...)` when it cannot find the credential, causing the CLI to exit immediately (hard fatal) instead of returning a controlled error.

## Steps to Reproduce
1. Run command: `go build -o harbor ./cmd/harbor`
2.  Edit your Harbor config so current-credential-name is set to a name that does not exist in the credentials list (or otherwise remove the named credential).
3. Run the password change: `./harbor password --old-password oldpass --new-password newpass`
4. Observe the process terminates with a fatal exit instead of returning a normal CLI error.

## Expected Behavior
The password command should return a descriptive error if the current credential cannot be found and exit gracefully with a non-zero code (no abrupt process termination).
Helper functions (e.g., `UpdateCredentialsInConfigFile`, `AddCredentialsToConfigFile`) should return errors instead of calling `log.Fatalf`, allowing callers to handle/report failures.

## Actual Behavior
`UpdatePassword` ignores the error from `GetCredentials` (`existingCred, _ := utils.GetCredentials(...)`) and builds a credential from zero-value fields.
`UpdateCredentialsInConfigFile` logs a fatal error (`log.Fatalf("credential with name '%s' not found", ...)`) and terminates the process immediately

## Environment
- OS: Linux
- Go: go1.24.8 linux/amd64
- Tool version: local build from this repository (no release tag)
- Repo/branch: local workspace (include commit/branch when filing remotely)

## Additional Context
This is triggered by stale or incorrect current-credential-name or by removing/mutating the named credential in the config.

madmecodes:
Fixes two bugs in the password command flow:

**GetCredentials error ignored (#742)**

`UpdatePassword` discarded the error from `GetCredentials` with `_`,
then built a credential from zero-value fields. When the credential
name didn't match anything, `UpdateCredentialsInConfigFile` called
`log.Fatalf` and killed the process instead of returning an error.

Fixed by:
- Checking the error from `GetCredentials` and returning early
- Replacing all `log.Fatalf` calls in `AddCredentialsToConfigFile` and
  `UpdateCredentialsInConfigFile` with `return fmt.Errorf(...)` so
  callers can handle failures normally

**User password reset has no flag support (#707)**

The `harbor user password` command had no flags registered, so `opts`
was always zero-valued. The only path was the interactive form.

Fixed by:
- Adding `--new-password` and `--confirm-password` flags
- When both flags are provided, validating and calling the API directly
- When flags are missing, falling back to the interactive form

Closes #707
Closes #742

### DIFF
--- a/cmd/harbor/root/user/password.go
+++ b/cmd/harbor/root/user/password.go
@@ -62,17 +62,16 @@ func UserPasswordChangeCmd() *cobra.Command {
 			var opts reset.PasswordChangeView
 
 			if passwordStdin {
-				fmt.Print("New Password: ")
+				fmt.Println("New Password: ")
 				passwordBytes, err := term.ReadPassword(int(os.Stdin.Fd())) // #nosec G115 - fd fits in int on all supported platforms
 				if err != nil {
 					return fmt.Errorf("failed to read password from stdin: %v", err)
 				}
-				fmt.Print("\nConfirm Password: ")
+				fmt.Println("Confirm Password: ")
 				confirmBytes, err := term.ReadPassword(int(os.Stdin.Fd())) // #nosec G115 - fd fits in int on all supported platforms
 				if err != nil {
 					return fmt.Errorf("failed to read confirmation password from stdin: %v", err)
 				}
-				fmt.Println()
 				if string(passwordBytes) != string(confirmBytes) {
 					return fmt.Errorf("passwords do not match")
 				}
--- a/cmd/harbor/root/user/password.go
+++ b/cmd/harbor/root/user/password.go
@@ -62,12 +62,20 @@ func UserPasswordChangeCmd() *cobra.Command {
 			var opts reset.PasswordChangeView
 
 			if passwordStdin {
-				fmt.Print("Password: ")
+				fmt.Print("New Password: ")
 				passwordBytes, err := term.ReadPassword(int(os.Stdin.Fd())) // #nosec G115 - fd fits in int on all supported platforms
 				if err != nil {
 					return fmt.Errorf("failed to read password from stdin: %v", err)
 				}
+				fmt.Print("\nConfirm Password: ")
+				confirmBytes, err := term.ReadPassword(int(os.Stdin.Fd())) // #nosec G115 - fd fits in int on all supported platforms
+				if err != nil {
+					return fmt.Errorf("failed to read confirmation password from stdin: %v", err)
+				}
 				fmt.Println()
+				if string(passwordBytes) != string(confirmBytes) {
+					return fmt.Errorf("passwords do not match")
+				}
 				opts.NewPassword = string(passwordBytes)
 				if err := utils.ValidatePassword(opts.NewPassword); err != nil {
 					return err
--- a/cmd/harbor/root/user/password.go
+++ b/cmd/harbor/root/user/password.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/goharbor/harbor-cli/pkg/api"
 	"github.com/goharbor/harbor-cli/pkg/prompt"
+	"github.com/goharbor/harbor-cli/pkg/utils"
 	"github.com/goharbor/harbor-cli/pkg/views/password/reset"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -36,10 +37,6 @@ func UserPasswordChangeCmd() *cobra.Command {
 			var userId int64
 			var err error
 			log.SetOutput(cmd.OutOrStderr())
-			resetView := &reset.PasswordChangeView{
-				NewPassword:     opts.NewPassword,
-				ConfirmPassword: opts.ConfirmPassword,
-			}
 
 			if len(args) > 0 {
 				userId, err = api.GetUsersIdByName(args[0])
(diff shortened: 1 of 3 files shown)


qcserestipy (changes requested):
The credential error fixes and `log.Fatalf` → `return fmt.Errorf` refactor all LGTM. A few issues with the flag approach though:

**Security: clear-text password flags are a bad idea but there's a precedent to follow.**
`--new-password`/`--confirm-password` expose the password in shell history, `ps` output, and system logs. The `harbor login` command already handles this correctly: it accepts `--password` for parity but also supports `--password-stdin`, which reads from stdin without echoing and avoids shell history. This command should follow the same pattern replace `--new-password` / `--confirm-password` with `--password-stdin` (using `term.ReadPassword` like login does), drop `--confirm-password` entirely since piped input doesn't need confirmation, and fall back to the TUI when the flag isn't set.

**Bug: providing only `--new-password` still opens the TUI prompt.**
The branch condition is `opts.NewPassword != "" && opts.ConfirmPassword != ""` so if only one flag is provided it silently falls through to the interactive form and ignores the flag value entirely. At minimum there should be an error if one flag is provided without the other.

**`--user-id` flag is missing.**
For a truly non-interactive flow (e.g. scripting/CI), a `--user-id` flag is also needed otherwise you still hit the interactive user-picker even when both password flags are set.

madmecodes:
Reworked per your feedback. Replaced --new-password/--confirm-password with --password-stdin using term.ReadPassword (same pattern as harbor login). Added --user-id flag for fully non-interactive CI flows. Falls back to TUI form when neither flag is set.

cmd/harbor/root/user/password.go
```diff
@@ -56,18 +59,39 @@ func UserPasswordChangeCmd() *cobra.Command {
 				}
 			}
 
-			reset.ChangePasswordView(resetView)
+			var opts reset.PasswordChangeView
 
-			err = api.ResetPassword(userId, *resetView)
+			if passwordStdin {
+				fmt.Print("Password: ")
+				passwordBytes, err := term.ReadPassword(int(os.Stdin.Fd())) // #nosec G115 - fd fits in int on all supported platforms
```
qcserestipy, line 66: A second confirm Password prompt might be great

cmd/harbor/root/user/password.go
```diff
@@ -56,18 +59,39 @@ func UserPasswordChangeCmd() *cobra.Command {
...
+				if err != nil {
+					return fmt.Errorf("failed to read password from stdin: %v", err)
+				}
+				fmt.Println()
```
qcserestipy: This can most likely be removed and added as \n newline to the password prompt

qcserestipy (changes requested):
Thank for applying the changes, just left a small comment. Additionally, the password reset with `--user-id` seems not to work:

```
➜  harbor-cli git:(pr/madmecodes/857) ./bin/harbor-cli user password --password-stdin --user-id 13197
Password: 
Error: failed to reset user password: [PUT /users/{user_id}/password][400] updateUserPasswordBadRequest 
```

madmecodes:
Added confirm password prompt for --password-stdin. Also merged the newline into the prompt string so there's no separate fmt.Println() between prompts.

On the 400 with --user-id: could you check whether the target user ID was your own account? The Harbor API requires old_password when changing your own password (even as admin), and this command only sends new_password since it's meant for admin resets of other users. If that's the case, `harbor password` is the right command for self-change.

If it was a different user's ID, I'd like to debug further -- could you share the password complexity you used? The client-side validation passes 8+ chars with upper/lower/digit, but the server might have stricter rules.

cmd/harbor/root/user/password.go
```diff
@@ -56,18 +59,47 @@ func UserPasswordChangeCmd() *cobra.Command {
...
+				if err != nil {
+					return fmt.Errorf("failed to read password from stdin: %v", err)
+				}
+				fmt.Print("\nConfirm Password: ")
+				confirmBytes, err := term.ReadPassword(int(os.Stdin.Fd())) // #nosec G115 - fd fits in int on all supported platforms
+				if err != nil {
+					return fmt.Errorf("failed to read confirmation password from stdin: %v", err)
+				}
+				fmt.Println()
+				if string(passwordBytes) != string(confirmBytes) {
```
qcserestipy: the Println is still here

cmd/harbor/root/user/password.go
```diff
@@ -56,18 +59,47 @@ func UserPasswordChangeCmd() *cobra.Command {
 				}
 			}
 
-			reset.ChangePasswordView(resetView)
+			var opts reset.PasswordChangeView
+
+			if passwordStdin {
+				fmt.Print("New Password: ")
```
qcserestipy: maybe just do fmt.Println(...)

cmd/harbor/root/user/password.go
```diff
@@ -56,18 +59,47 @@ func UserPasswordChangeCmd() *cobra.Command {
...
+				passwordBytes, err := term.ReadPassword(int(os.Stdin.Fd())) // #nosec G115 - fd fits in int on all supported platforms
+				if err != nil {
+					return fmt.Errorf("failed to read password from stdin: %v", err)
+				}
+				fmt.Print("\nConfirm Password: ")
```
qcserestipy: maybe just do fmt.Println(...)

qcserestipy (changes requested):
Thank you for the changes, funcitonality looks good to me. Just please clean up the Password prompts. 

What I found that could be covered in a second PR after this one was merged:

If I try to change the password of a user to the same password I get an error form the harbor api:

Error: failed to reset user password: [PUT /users/{user_id}/password][400] updateUserPasswordBadRequest 

We could handle this by double checking whether existing and new password are the same and throw a better error.

PR #875, opened 2026-05-05
feat: improve storage limit validation and unit support

### LINKED ISSUE #873: [bug]:  storage limit validation is currently inconsistent and contains logic bugs
## Description
The Harbor CLI's storage limit validation is currently restrictive and contains logic bugs that prevent users from using standard units or limits exceeding 1024 bytes.

- Logic Bug: The validation helper contains an impossible condition (storageLimit > -1 && storageLimit < 0).
- Incorrect Limit Check: The check storageLimit > 1024 incorrectly treats the value as bytes instead of terabytes, effectively capping storage at 1KB.
- Missing Error Handling: API handlers like CreateProject ignore parsing errors, leading to silent failures where invalid inputs are treated as 0.

## Steps to Reproduce
1. Run command: `harbor project create`
2. With arguments/config: `Interactive mode, enter 10GB for the storage limit.`
3. Observe error the storage limit only takes integer values

## Expected Behavior
The CLI should support standard binary units (MiB, GiB, TiB) and correctly convert them to bytes. In interactive mode, a unit selection field should be provided. A value of -1 should be accepted for unlimited storage.

## Actual Behavior
The CLI only accepts plain integers and fails for any string containing units. Additionally, any value greater than 1024 (bytes) is rejected due to an incorrect hardcoded limit check in the validator.

## Environment
- OS: any
- Tool version:harbor-cli (latest main)
- Other relevant details: go 1.2.1+

## Additional Context

Image

Yashbhu:
## Description
This PR enhances the Harbor CLI by supporting human-readable units (e.g., 10GB, 500MiB) in storage limit fields for projects and quotas. It also fixes several logic bugs in the validation helpers that caused valid limits to be rejected.

- Fixes #873 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [x] Refactor

## Changes
-  Updated `StorageStringToBytes` to distinctly support both **Decimal** (KB, MB, GB, TB = 1000^n) and **Binary** (KiB, MiB, GiB, TiB = 1024^n) units, matching industry standards.
-  Implemented defensive guards to prevent `int64` overflows when parsing extremely large storage strings.
-  Ensured a consistent maximum storage cap of **1024 TiB** is enforced across all input formats, including plain numeric byte values.
-  Fixed a bug in `ValidateStorageLimit` where storage was incorrectly capped at 1024 bytes instead of 1024 terabytes, and removed mathematically impossible validation conditions.
-  Updated `CreateProject` and `UpdateQuota` handlers to properly handle and return parsing errors instead of failing silently.
-  Added 15+ new unit test cases covering decimal/binary differentiation, edge cases, and overflow scenarios.

### DIFF
--- a/cmd/harbor/root/quota/update.go
+++ b/cmd/harbor/root/quota/update.go
@@ -55,13 +55,9 @@ func UpdateQuotaCommand() *cobra.Command {
 			}
 
 			if storage != "" {
-				if storage == "-1" {
-					storageValue = -1
-				} else {
-					storageValue, err = utils.StorageStringToBytes(storage)
-					if err != nil {
-						return fmt.Errorf("failed to parse storage: %v", err)
-					}
+				storageValue, err = utils.StorageStringToBytes(storage)
+				if err != nil {
+					return fmt.Errorf("failed to parse storage: %v", err)
 				}
 			} else {
 				storage = update.UpdateQuotaView(quota)
--- a/pkg/api/project_handler.go
+++ b/pkg/api/project_handler.go
@@ -26,20 +26,21 @@ import (
 )
 
 func CreateProject(opts create.CreateView) error {
+	storageLimit, err := utils.StorageStringToBytes(opts.StorageLimit)
+	if err != nil {
+		return err
+	}
+
 	ctx, client, err := utils.ContextWithClient()
 	if err != nil {
 		return err
 	}
+
 	registryID := new(int64)
 	*registryID, _ = strconv.ParseInt(opts.RegistryID, 10, 64)
 
 	if !opts.ProxyCache {
 		registryID = nil
-	}
-
-	storageLimit, err := utils.StorageStringToBytes(opts.StorageLimit)
-	if err != nil {
-		return err
 	}
 
 	public := strconv.FormatBool(opts.Public)
--- a/pkg/api/project_handler.go
+++ b/pkg/api/project_handler.go
@@ -37,7 +37,10 @@ func CreateProject(opts create.CreateView) error {
 		registryID = nil
 	}
 
-	storageLimit, _ := strconv.ParseInt(opts.StorageLimit, 10, 64)
+	storageLimit, err := utils.StorageStringToBytes(opts.StorageLimit)
+	if err != nil {
+		return err
+	}
 
 	public := strconv.FormatBool(opts.Public)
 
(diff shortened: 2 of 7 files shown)


Yashbhu:
@vg006 
can you PTAL

vg006 (changes requested):
Hello @Yashbhu, I have reviewed your changes, and these are my concerns.

Let the units be `MiB`, `GiB` and `TiB` and suggest not to change them to `MIB`, `GIB`, `TIB`, since,
- Users are familiar with the former units and CI scripts may contain hardcode storage limit units in the former pattern, so this change will break the flow, *though we have modified regexp pattern*
- Also the tool logs and prints in the former way `MiB`, `GiB`, `TiB`, but comparing internally as `MIB`, `GIB`, `TIB` doesn't make sense, we can stick with what already being followed.

Yashbhu:
@vg006 can you PTAL now

Yashbhu:
@vg006 friendly ping

vg006:
Yeah, LGTM, but I guess the `multiplier` map can be defined as the package-level private variable in `utils.go` file and the same can be used in the `helpers.go` file, if we move the validation logic from the `StorageStringToBytes` function to `ValidateStorageLimit` function.

What's your opinion on that? @Yashbhu 
cc: @bupd

Yashbhu:
> Yeah, LGTM, but I guess the `multiplier` map can be defined as the package-level private variable in `utils.go` file and the same can be used in the `helpers.go` file, if we move the validation logic from the `StorageStringToBytes` function to `ValidateStorageLimit` function.
> 
> What's your opinion on that? @Yashbhu cc: @bupd

great suggestion I've implemented this refactor by moving the multiplier map to a package-level private variable in utils.go and centralizing the validation logic in ValidateStorageLimit. I've pushed the changes to the branch.

Yashbhu:
@vg006 cc: @bupd PTAL

vg006:
Yeah I reviewed your changes. Actually, I was asking your opinion on the refactor and not wanted to perform it. I have discussed it with the maintainers in the last community meeting, but there is no significant or valid reason to refactor it. So let it be how it was in this [commit](https://github.com/goharbor/harbor-cli/pull/875/changes/d60100fa33ebbad1cf24fd4b9ccb49d99aaa79da). So it LGTM.

Also I'm a contributor too, reviewing PRs on behalf of maintainers, to help them to manage all. Since the maintainers are looking on it, I will leave it to them. Thank you for the contribution.

Yashbhu:
@bupd hi can you please open this branch?

PR #880, opened 2026-05-05
feat: allow single-label hostnames in ValidateURL and improve performance

### LINKED ISSUE #879: [bug]:The ValidateURL function fails to validate URLs that use single-label hostnames
## Description
The ValidateURL function fails to validate URLs that use single-label hostnames (e.g., http://harbor). This prevents the CLI from being used in internal networks, Kubernetes services, or Docker Compose environments where FQDNs are not required.

## Steps to Reproduce
1. Run command: `harbor login harbor`
2. With arguments/config: `Use a single-label hostname (no dots) as the server URL.`
3. Observe error nvalid URL format: invalid host: must be a valid IP address or domain name

## Expected Behavior
The CLI should successfully parse and validate single-label hostnames according to RFC 1123, allowing connection to internal Harbor instances.

## Actual Behavior
The validator rejects any hostname that does not contain a dot and a TLD of at least 2 characters.

## Environment
- OS: macOS / Linux / Windows
- Tool version: harbor-cli (latest)
- Other relevant details: Docker/Kubernetes internal networking.

## Additional Context
Logs, screenshots, or other useful information.

Yashbhu:
## Description
This PR relaxes the URL validation logic to support internal hostnames (single-label hostnames like `http://harbor`), which are standard in DevOps environments but were previously blocked. It also improves performance by optimizing regex handling.

- Fixes #879 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
- **RFC 1123 Support**: Updated the domain regex to allow hostnames without dots, enabling compatibility with Kubernetes service names and Docker Compose networking.
- **Performance Optimization**: Moved the `domainNameRegex` to a package-level global variable to avoid the overhead of re-compiling the regex on every `ValidateURL` call.
- **Enhanced Testing**: Added new unit test cases for internal service names (e.g., `http://harbor`, `https://registry-server`) and verified they pass alongside existing FQDN and IP tests.

### DIFF
--- a/pkg/utils/helper.go
+++ b/pkg/utils/helper.go
@@ -25,6 +25,16 @@ import (
 	"unicode"
 )
 
+var (
+	emailRegex       = regexp.MustCompile(`^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$`)
+	configPathRegex  = regexp.MustCompile(`^[\w./-]{1,255}\.(yaml|yml)$`)
+	flRegex          = regexp.MustCompile(`^[A-Za-z]{1,20}\s[A-Za-z]{1,20}$`)
+	tagNameRegex     = regexp.MustCompile(`^[\w][\w.-]{0,127}$`)
+	projectNameRegex = regexp.MustCompile(`^[a-z0-9][a-z0-9._-]{0,254}$`)
+	registryNameRegex = regexp.MustCompile(`^[\w][\w.-]{0,63}$`)
+	domainNameRegex  = regexp.MustCompile(`^(?i:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?)(?:\.[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?)*$`)
+)
+
 
 func pluralise(value int, unit string) string {
 	if value == 1 {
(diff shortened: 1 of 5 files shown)


Yashbhu:
i did some changes in the helper_test.go:

http://harbor: It verifies that a simple, single-word hostname (which used to fail) now passes.
https://registry-server: This verifies that the regex correctly handles hyphens inside a single label. Many Kubernetes services use hyphens (like harbor-core), so i think we need to make sure they work
http://harbor:8080: This verifies that the hostname validation still works when a port number is attached to it we want to make sure our regex only looks at the host part and doesn't get confused by the :8080.

Yashbhu:
@bupd  PTAL whenever you have bandwidth 
if this issue makes sense or not https://github.com/goharbor/harbor-cli/issues/879/
thanks !

pkg/utils/helper.go
```diff
@@ -177,11 +183,9 @@ func ValidateRegistryName(rn string) bool {
 	return re.MatchString(rn)
 }
 
-// ValidateURL checks if the URL has valid format, non-empty host, and host is a valid IP or domain.
-// Domain regex: labels must start/end with alphanumeric, can contain hyphens, max 63 chars, TLD min 2 letters.
+// ValidateURL checks that the URL has a valid format and a non-empty host.
+// The host may be an IP address, "localhost", or an RFC 1123-style hostname validated by domainNameRegex.
 func ValidateURL(rawURL string) error {
-	var domainNameRegex = regexp.MustCompile(`^(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,}$`)
```
bupd, line 183: I dont know why are you moving removing the var from here - just to declare above the function. 

That doesnt make sense - is there any specific reason to do this. 

thanks

bupd (changes requested):

pkg/utils/helper.go
Yashbhu: the only reason to move it outside i thought of is to avoid the overhead of re-compiling the regex on every call
though you are right to address it that in a CLI it's not much needed and rather clean code setup should be much prioritized i've reverted this 
thanks !

Yashbhu:
@bupd PTAL

PR #901, opened 2026-05-08
 Fix: make harbor registry update non-interactive when flags are provided

### LINKED ISSUE #898: [bug]: Missing non-interactive registry update mode
## Description
I'm unable to update a registry non-interative.

## Steps to Reproduce
```shell
harbor-cli registry update docker.muh --name "docker.muh" --url "https://docker.io" --description "Docker Muh" --credential-access-key "" --credential-access-secret ""
```

## Expected Behavior
Only defined fields should be updated without asking.

## Actual Behavior
Provides an interactive input mask.

## Environment
- OS: Rocky Linux 9.7
- Tool version: 0.0.19
- Other relevant details:

## Additional Context
Logs, screenshots, or other useful information.

Image

rakshityadav1868:
## Description
The `harbor registry update` command incorrectly opens interactive prompts even when all update values are explicitly passed through CLI flags.

- Fixes #898 

## Type of Change
- [x] Bug fix

# Changes Made

####  File: `cmd/harbor/root/registry/update.go`

Untitled design(2)


NucleoFusion (changes requested):
I dont think this is the best way to do it, 
please refer to other commands where we do the same, and how it happens there.

rakshityadav1868:
@NucleoFusion 
Updated the implementation to follow the existing pattern used in instance/update.go by moving the flag change detection into a helper function instead of keeping the inline condition.

cmd/harbor/root/registry/update.go
```diff
@@ -118,3 +121,15 @@ func UpdateRegistryCommand() *cobra.Command {
...
+	flags := cmd.Flags()
+	return flags.Changed("name") ||
+		flags.Changed("type") ||
+		flags.Changed("description") ||
+		flags.Changed("url") ||
+		flags.Changed("insecure") ||
+		flags.Changed("credential-access-key") ||
+		flags.Changed("credential-access-secret") ||
+		flags.Changed("credential-type")
+}
```
NucleoFusion, line 142: I meant that rather than checking flags, it would be better to check the values directly of `opts`

NucleoFusion (changes requested):

rakshityadav1868:
@NucleoFusion 
Done instead of flags used opts

NucleoFusion (commented):
Okay so, in the case of this command, we were not checking flags at all
So,
Traditionally, we check the minimal flags required for the command to execute.
From the API Docs, there is no specific amount of flags that are either required or not required.

For this, it would be best to discuss in the meeting what amount of flags from the registry model

```go
{
  "name": "string",
  "description": "string",
  "url": "string",
  "credential_type": "string",
  "access_key": "string",
  "access_secret": "string",
  "insecure": true
}
```
we consider to be required, and what we dont.
This may even end up amending how we currently process the model

rakshityadav1868:
@NucleoFusion updated the flow now for updating registry through flag requires Id as a mandatory field!!

https://github.com/user-attachments/assets/609c755b-073c-4114-a8d8-9f09acb5a308

bupd (changes requested):
No, Instead of making ID mandatory i would do a roundtrip with just name to get the id if it fails i would say name is not found. and throw err msg to user. 

also the direction this pr is heading is not consistent with other cmds present in the harbor-cli. I suggest @rakshityadav1868 to reconsider the implementation with how other cmds treat this behaviour before implementing. 

keeping things consistent across pkgs is more important than a random fix.

rakshityadav1868:
Hey @bupd,

I've addressed the review feedback and aligned the implementation with the existing Harbor CLI update patterns.

### Changes made

* Replaced the inline update flag handling with `hasUpdateFlagChanges()` using `flags.Changed(...)` to determine interactive vs. non-interactive mode.
* Moved the flag application logic into `applyUpdateFlags()` to keep `RunE` cleaner and consistent with other update commands.
* Preserved the existing interactive TUI flow when no update flags are provided.
* Non-interactive updates now skip the TUI and update only the explicitly provided fields while preserving the remaining registry configuration.
* Continued using the existing registry name to ID lookup (`api.GetRegistryIdByName`) to stay consistent with other registry commands.

rakshityadav1868:
https://github.com/user-attachments/assets/fe1e5633-510f-4bf4-ae67-3c32cb131c6c

PR #904, opened 2026-05-08
Feat: support --registry-name for proxy cache project creation

### LINKED ISSUE #900: [feature]: Create project with --registry-name instead of --registry-id
## Problem
For automation purposes it would be nice to pass `--registry-name` like:
```shell
harbor-cli project create docker.io --proxy-cache --public --registry-name "docker.io"  --storage-limit "-1"
```
Instead of:
```shell
harbor-cli registy list --output-format json
# parsing id
harbor-cli project create docker.io --proxy-cache --public --registry-id 1  --storage-limit "-1"
```

rakshityadav1868:
## Description
Previously, users had to manually list registries, parse the registry ID, and pass `--registry-id` during project creation. With this update, users can now directly provide the registry name and the CLI will internally resolve it to the corresponding registry ID before sending the request to Harbor. 
This keeps the existing Harbor architecture unchanged, since Harbor still internally operates using `registry_id`.

- Fixes #900 

- [x] New feature

### 1. Added `--registry-name` flag

Added support for a new CLI flag:

- `--registry-name`
### 2. Added registry name to ID resolution

Added logic to resolve registry names using:

- `api.GetRegistryIdByName(..)`
### 3. Updated validation logic

Updated proxy cache validation to support both:

- `--registry-id`

- `--registry-name`
---

Untitled design(3)

---------------
-------
Untitled design(4)

### DIFF
--- a/cmd/harbor/root/project/create.go
+++ b/cmd/harbor/root/project/create.go
@@ -26,6 +26,7 @@ import (
 // CreateProjectCommand creates a new `harbor create project` command
 func CreateProjectCommand() *cobra.Command {
 	var opts create.CreateView
+	var registryName string
 
 	cmd := &cobra.Command{
 		Use:   "create [project name]",
@@ -37,9 +38,19 @@ func CreateProjectCommand() *cobra.Command {
 			if len(args) > 0 {
 				opts.ProjectName = args[0]
 			}
+			if registryName != "" {
+				registryID, err := api.GetRegistryIdByName(registryName)
+				if err != nil {
+					return fmt.Errorf("failed to get registry ID for name %s: %v", registryName, err)
+				}
+				if registryID == 0 {
+					return fmt.Errorf("registry with name %s not found", registryName)
+				}
+				opts.RegistryID = fmt.Sprintf("%d", registryID)
+			}
 
 			if opts.ProxyCache && opts.RegistryID == "" {
-				return fmt.Errorf("proxy cache selected but no registry ID provided. Use --registry-id")
+				return fmt.Errorf("proxy cache selected but no registry ID or registry name provided. Use --registry-id or --registry-name")
 			}
 
 			if !opts.ProxyCache && opts.RegistryID != "" {
@@ -75,6 +86,7 @@ func CreateProjectCommand() *cobra.Command {
 	flags := cmd.Flags()
 	flags.BoolVarP(&opts.Public, "public", "", false, "Project is public or private")
 	flags.StringVarP(&opts.RegistryID, "registry-id", "", "", "ID of referenced registry when creating the proxy cache project")
+	flags.StringVarP(&registryName, "registry-name", "", "", "Name of referenced registry when creating the proxy cache project (alternative to --registry-id)")
 	flags.StringVarP(&opts.StorageLimit, "storage-limit", "", "", "Storage quota of the project")
 	flags.BoolVarP(&opts.ProxyCache, "proxy-cache", "", false, "Whether the project is a proxy cache project")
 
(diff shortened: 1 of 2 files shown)


cmd/harbor/root/project/create.go
```diff
@@ -75,6 +86,7 @@ func CreateProjectCommand() *cobra.Command {
 	flags := cmd.Flags()
 	flags.BoolVarP(&opts.Public, "public", "", false, "Project is public or private")
 	flags.StringVarP(&opts.RegistryID, "registry-id", "", "", "ID of referenced registry when creating the proxy cache project")
+	flags.StringVarP(&registryName, "registry-name", "", "", "Name of referenced registry when creating the proxy cache project (alternative to --registry-id)")
```
NucleoFusion: I think we can drop the `(alternative to --registry-id)` since that makes the string too long, and doesnt really add any value, imo

cmd/harbor/root/project/create.go
```diff
@@ -37,9 +38,19 @@ func CreateProjectCommand() *cobra.Command {
...
+			if registryName != "" {
+				registryID, err := api.GetRegistryIdByName(registryName)
+				if err != nil {
+					return fmt.Errorf("failed to get registry ID for name %s: %v", registryName, err)
+				}
+				if registryID == 0 {
+					return fmt.Errorf("registry with name %s not found", registryName)
+				}
+				opts.RegistryID = fmt.Sprintf("%d", registryID)
+			}
```
NucleoFusion, line 54: What is user provides both registry-id and registry-name
Consider throwing an error when both provided.

NucleoFusion (changes requested):

rakshityadav1868:
@NucleoFusion
 Shorten the string and added a condition when both name and id is provided it throws an error.
 
Screenshot 2026-05-09 at 12 21 35 PM

NucleoFusion (changes requested):
The implementation works, yes
But could you also modify the view to accommodate this?
Right now, if we only provide the --registry-name the view opens (as it should) but it does not auto-select the ID which was provided.

### Steps to Reproduce

1. Run ` harbor-cli project create --registry-name dockerhub --proxy-cache`
2. Fill the Initial details, and move to second page.
ERR: In the second page, dockerhub is not the one selected.

rakshityadav1868:
@NucleoFusion i think in this Pr it is fixed i have tried on this Pr branch cant find the issue
https://github.com/user-attachments/assets/e623bbda-b237-4ef1-b64a-425e1b097b86

NucleoFusion:
> @NucleoFusion i think in this Pr it is fixed i have tried on this Pr branch cant find the issue [github.com/user-attachments/assets/e623bbda-b237-4ef1-b64a-425e1b097b86](https://github.com/user-attachments/assets/e623bbda-b237-4ef1-b64a-425e1b097b86)

Try some other registry
broadcom or somethign maybe.
There are a lot of registries, its only showing 2, thats not correct

rakshityadav1868:
@NucleoFusion 
means under registryId all registries should be shown ? right ?

Famous077:
Hi @rakshityadav1868 , Before raising any PR, you have to ask is there anyone working on this issue or not? Your activity is totally against the Open source contributing guideline.

rakshityadav1868:
Hi @NucleoFusion 
Implemented the changes with conditional form groups instead of WithHideFunc.

Reason for not using WithHideFunc:

* The select component was still being rendered internally, even when hidden
* That was causing the Huh viewport/rendering issue with registry selection

Current approach:

* The registry select group is only appended when required
* If --registry-id or --registry-name is already provided, the select component is not rendered at all

https://github.com/user-attachments/assets/34e9c77d-5e6c-40e4-ae3c-4a0780e43c9a

rakshityadav1868:
One small clarification currently if neither --registry-id nor --registry-name is provided with --proxy-cache, it already throws an error before reaching the form flow (this behavior existed before my changes as well).

mreiche:
Any progress on this?

PR #909, opened 2026-05-09
fix: resolve project member update by username and sync update/create wrappers

dev-aditya-hub:
## Summary

Found three sibling-symmetry bugs while looking around the member, scanner, and webhook handlers — in each case a fix or normalization exists on the create path but was never replicated on the update path, so an update silently produces a wrong result.

### `project member update --member <name>` was targeting the wrong row

`cmd/harbor/root/project/member/update.go` was resolving `--member <name>` via `api.GetUsersIdByName`, which returns a **user ID**, and passing it as `Mid` into `UpdateProjectMember` — but Harbor's API expects the **project member ID** (`project_members.id`). These are different rows and only line up by accident.

Introduced `api.UpdateMemberByUsername` (mirroring the existing `DeleteMemberByUsername`) which calls `ListMembers` and matches on `EntityName`, then routes the cmd through it. The numeric `[memberName]` positional and the interactive picker still use the existing `UpdateMember` path.

### `scanner update --auth None` was leaving the scanner with unrecognized auth

`pkg/api/scanner_handler.go::CreateScanner` normalizes `\"None\" -> \"\"`; `UpdateScanner` did not. Both the `--auth` flag and the interactive view surface the literal string `\"None\"`, so an update would persist that as the auth method and the scanner adapter could no longer authenticate. Added the same normalization to `UpdateScanner`.

### `webhook update` was not trimming the endpoint URL

`pkg/api/webhook_handler.go::CreateWebhook` calls `strings.TrimSpace` on the endpoint URL; `UpdateWebhook` did not. A trailing newline/space pasted into `--endpoint-url` would be stored verbatim. Added the trim.

## Test plan

- [ ] `harbor project member update <p> --member <user> --roleid 2` updates the role of `<user>` (and not whichever member happens to share that user's ID)
- [ ] `harbor project member update <p> <memberID> --roleid 2` still works (positional path unchanged)
- [ ] `harbor scanner update <name> --auth None` no longer leaves the scanner with `Auth: \"None\"` in Harbor
- [ ] `harbor webhook update ... --endpoint-url \"https://example.com/hook \"` stores the URL without trailing whitespace

### DIFF
--- a/cmd/harbor/root/project/member/update.go
+++ b/cmd/harbor/root/project/member/update.go
@@ -53,36 +53,36 @@ func UpdateMemberCommand() *cobra.Command {
 				}
 			}
 
-			if memberName == "" {
-				opts.ID = prompt.GetMemberIDFromUser(opts.ProjectNameOrID, memberName)
-
-				if opts.ID == 0 {
-					return fmt.Errorf("No members found in project")
-				}
-			} else {
-				opts.ID, err = api.GetUsersIdByName(memberName)
-				if err != nil {
-					return err
-				}
-			}
-
 			if roleID == 0 {
 				roleID = prompt.GetRoleIDFromUser()
 			}
-			opts.RoleID = &models.RoleRequest{
-				RoleID: roleID,
-			}
+			role := &models.RoleRequest{RoleID: roleID}
 
 			// when set true parses projectNameOrID as projectName
 			// else it parses as an integer ID
 			opts.XIsResourceName = !isID
+			opts.RoleID = role
 
-			err = api.UpdateMember(opts)
-			if err != nil {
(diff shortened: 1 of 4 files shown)


dev-aditya-hub:
@qcserestipy gentle ping — would appreciate a look here when you have a slot.

This one is a follow-up sweep after #908: same shape of bug (a fix lives on the create path, the matching update path was missed), three different commands. The biggest of the three is the `member update --member <name>` path — it was passing a user ID where Harbor expects a project-member row ID, so on any non-toy tenant the role change lands on the wrong member while the CLI still prints success.

qcserestipy:
Since there was no feedback

PR #914, opened 2026-05-10
fix: remove password policy check from login

### LINKED ISSUE #911: [bug]: password policy enforcement breaks login
## Description
Client-side password policy enforcement breaks login functionality. I have an autogenerated (by the `harbor-helm` helm chart) password for the admin account. I can login without problems via WebUI, but attempting setup using `harbor login` causes "* wrong! the password doesn't have an uppercase letter" message 

## Steps to Reproduce
1. Run command: `harbor login`
2. With password non-conforming to password policy (in my case lowercase alphanumeric)
3. Observe error

## Expected Behavior
Client should not prevent login, because of violated password policy.

## Actual Behavior
Password policy prevents login, even though the password is correct

## Environment
- OS: Arch Linux
- Tool version: v0.0.19
- Other relevant details: harbor instance deployed via `harbor-helm` helmchart

madmecodes:
## Summary

- Removes `ValidatePassword()` call from the interactive login form's password field
- Password policy enforcement remains in place for user creation, password change, and password reset flows where it belongs
- Fixes the case where auto-generated passwords (e.g. from harbor-helm) get rejected at login even though the server accepts them

The login form still validates that the password is non-empty -- it just no longer enforces format rules like uppercase/lowercase/digit requirements, which is purely a creation-time concern.

Fixes #911

### DIFF
--- a/pkg/views/login/create.go
+++ b/pkg/views/login/create.go
@@ -68,9 +68,6 @@ func CreateView(loginView *LoginView) {
 					if strings.TrimSpace(str) == "" {
 						return errors.New("password cannot be empty or only spaces")
 					}
-					if err := utils.ValidatePassword(str); err != nil {
-						return err
-					}
 					return nil
 				}),
 			huh.NewInput().


qcserestipy:
Since this is a upstream harbor issue and not a cli issue this is not the way we should proceed, but wait for a solution from upstream

madmecodes:
makes sense, closing this one. thanks for the heads up!

PR #930, opened 2026-05-15
fix(config): prevent silent success on empty configs and fix context switch error propagation

### LINKED ISSUE #929: [bug]: Silent success on empty configurations and improper context switch error propagation
## Description
Currently, when a user applies an empty or invalid configuration file using the CLI, the operation exits silently as a "success" instead of alerting the user. Additionally, failures during context switching print to `stdout` but do not return proper exit codes, which breaks automation and scripting workflows.

## Steps to Reproduce
1. Run command: `harbor configurations apply -f empty-config.yaml`
2. Observe error: The CLI exits with a `0` success code without warning that nothing was applied.
3. Try switching to an invalid context using `harbor context switch`.
4. Observe error: The CLI prints a failure message but still exits with a `0` code.

## Expected Behavior
The CLI should return an explicit error and exit with a non-zero exit code (`1`) when configuration files are completely empty/nil or when context switching fails.

## Actual Behavior
The CLI exits gracefully (`0`) making it seem like the operations succeeded.

## Environment
- OS: All
- Tool version: Latest main
- Other relevant details: Affects `cmd/harbor/root/configurations/apply.go` and `cmd/harbor/root/context/switch.go`.

gcharpe1604:
## Description
This PR addresses ungraceful handling of empty configuration files and context switching errors. Previously, applying an empty configuration resulted in a silent success (`exit 0`), and failing to switch contexts printed an error but still exited gracefully, which breaks automation scripts.

- Fixes #929 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
- Added nil checks in `cmd/harbor/root/configurations/apply.go` to error out on empty configuration files instead of silently succeeding.
- Refactored `cmd/harbor/root/context/switch.go` to return an explicit error, allowing Cobra to emit a proper non-zero exit code (`exit 1`) on failure.

### DIFF
--- a/cmd/harbor/root/configurations/apply.go
+++ b/cmd/harbor/root/configurations/apply.go
@@ -66,8 +66,7 @@ Make sure to run 'harbor config get' first to populate the local config file wit
 				}
 
 				if configurations == nil {
-					warningStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("3")).Bold(true)
-					fmt.Println(warningStyle.Render("Warning: Config file is empty or contains no valid configurations."))
+					fmt.Println("Warning: Config file is empty or contains no valid configurations.")
 					return nil
 				}
 			} else {
--- a/cmd/harbor/root/configurations/apply.go
+++ b/cmd/harbor/root/configurations/apply.go
@@ -66,7 +66,9 @@ Make sure to run 'harbor config get' first to populate the local config file wit
 				}
 
 				if configurations == nil {
-					return fmt.Errorf("config file is empty or contains no valid configurations")
+					warningStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("3")).Bold(true)
+					fmt.Println(warningStyle.Render("Warning: Config file is empty or contains no valid configurations."))
+					return nil
 				}
 			} else {
 				return fmt.Errorf("no config file specified")
--- a/cmd/harbor/root/configurations/apply.go
+++ b/cmd/harbor/root/configurations/apply.go
@@ -64,6 +64,10 @@ Make sure to run 'harbor config get' first to populate the local config file wit
 				default:
 					return fmt.Errorf("unsupported file type: %s, expected '.yaml/.yml' or '.json'", fileType)
 				}
+
+				if configurations == nil {
+					return fmt.Errorf("config file is empty or contains no valid configurations")
+				}
 			} else {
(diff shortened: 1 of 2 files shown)


gcharpe1604:
@qcserestipy @NucleoFusion 

Would appreciate your feedback/review on this PR whenever you have time. Thanks!

cmd/harbor/root/configurations/apply.go
```diff
@@ -64,6 +64,10 @@ Make sure to run 'harbor config get' first to populate the local config file wit
 				default:
 					return fmt.Errorf("unsupported file type: %s, expected '.yaml/.yml' or '.json'", fileType)
 				}
+
+				if configurations == nil {
+					return fmt.Errorf("config file is empty or contains no valid configurations")
```
qcserestipy: This should be no error, this should be a warning. In principle applying empty config is not harmful therefore no error is needed

qcserestipy (changes requested):

cmd/harbor/root/configurations/apply.go
```diff
@@ -64,6 +64,12 @@ Make sure to run 'harbor config get' first to populate the local config file wit
 				default:
 					return fmt.Errorf("unsupported file type: %s, expected '.yaml/.yml' or '.json'", fileType)
 				}
+
+				if configurations == nil {
+					warningStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("3")).Bold(true)
```
qcserestipy: Currently, we do not enforce styling for warnings, a simple fmt.Println should be enough

qcserestipy (changes requested):
Thanls for the applied changes, please see my comment

cmd/harbor/root/configurations/apply.go
gcharpe1604: Switched to standard `fmt.Println`.

gcharpe1604:
@qcserestipy Thanks for the review!!

I've addressed the latest feedback and pushed an update. Please let me know if there are any remaining concerns.

qcserestipy (approved):
LGTM

gcharpe1604:
@qcserestipy Thanks for the approval!!

bupd (approved):
/lgtm

PR #932, opened 2026-05-15
fix(replication): propagate interactive prompt errors to prevent abrupt CLI crashes

### LINKED ISSUE #931: [bug]: Interactive prompt helpers and views crash abruptly or panic on API errors and user aborts
## Description

The replication interactive prompt helpers (`GetReplicationPolicyFromUser`, `GetReplicationExecutionIDFromUser`, and `GetReplicationTaskIDFromUser`) use `log.Fatal()` internally when API calls fail or return empty payloads. Additionally, some replication selection views terminate the process directly using `os.Exit(1)`.

These abrupt exits bypass Cobra's `RunE` error handling mechanism and prevent commands from returning errors gracefully to the caller.

## Steps to Reproduce

1. Run a replication command that triggers an interactive selection (for example, `harbor replication start` without providing an ID).
2. Trigger an API failure or use an environment where no selectable resources exist.
3. Observe that the CLI terminates abruptly instead of returning an error through Cobra.

## Expected Behavior

Prompt helpers and views should return errors back to the caller command. Cobra's `RunE` handlers should propagate those errors and exit cleanly with a standard non-zero exit code.

## Actual Behavior

The application terminates immediately via `log.Fatal()` or `os.Exit(1)` instead of returning an error to the command layer.

## Environment

* OS: All
* Tool version: Latest main
* Other relevant details: Affects replication prompt helpers in `pkg/prompt/prompt.go` and related replication selection views and commands.

gcharpe1604:
# Description

This PR fixes replication interactive prompt helpers and views that terminate the CLI abruptly via `log.Fatal()` or `os.Exit(1)` during API failures or invalid selections.

Errors are now propagated through Cobra's command lifecycle (`RunE`) instead of terminating the process directly, resulting in more consistent and automation-friendly CLI behavior.

Fixes #931

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactor (non-breaking change which improves code structure)
- [ ] Documentation update
- [ ] Chore / maintenance

### 1. Removal of Abrupt CLI Crashes (`log.Fatal()`, `os.Exit()`)
* **Prompt Helpers:** Removed all instances of `log.Fatal()` in prompt helper functions in `pkg/prompt/prompt.go` when replication APIs fail or return empty lists.
* **View Tier:** Replaced direct `os.Exit(1)` calls inside selection views (such as `ReplicationExecutionList` and `ReplicationTasksList`) with clean error propagation via channels.
* **Cobra Integration:** Errors are returned up to Cobra's command layer, allowing it to format error messages and return appropriate exit codes.

### 2. Error & Cancel/Abort Propagation
* Added an `errChan chan<- error` channel to the asynchronous view functions to propagate errors back to the caller prompts.
* If a selection is cancelled/aborted (yielding empty selection choice), an error is sent to the channel instead of calling `os.Exit(1)`.

## How to Test

1. **Verify Cancel/Abort Graceful Exit:**
   Run an interactive command like `harbor replication start` without arguments, cancel the prompt selection, and verify the command exits gracefully with `user aborted selection` instead of calling `log.Fatal()`, calling `os.Exit(1)`, or silently executing on default values.

2. **Replication Commands:**
   Verify `harbor replication start`, `harbor replication stop`, and `harbor replication log` function as expected and terminate cleanly if aborted.

### DIFF
--- a/pkg/views/replication/execution/select/view.go
+++ b/pkg/views/replication/execution/select/view.go
@@ -25,7 +25,7 @@ import (
 
 var ErrUserAborted = errors.New("user aborted selection")
 
-func ReplicationExecutionList(executions []*models.ReplicationExecution) (int64, error) {
+func ReplicationExecutionList(executions []*models.ReplicationExecution, choice chan<- int64, errChan chan<- error) {
 	itemsList := make([]list.Item, len(executions))
 	for i, p := range executions {
 		displayName := fmt.Sprintf("ID: %d, Status: %s, Trigger: %s, Start Time: %s, Succeed: %d, Total: %d",
@@ -37,24 +37,25 @@ func ReplicationExecutionList(executions []*models.ReplicationExecution) (int64,
 
 	p, err := tea.NewProgram(m, tea.WithAltScreen()).Run()
 	if err != nil {
-		return 0, fmt.Errorf("error running selection program: %w", err)
+		errChan <- fmt.Errorf("error running selection program: %w", err)
+		return
 	}
 
 	if model, ok := p.(selection.Model); ok {
-		if model.Aborted {
-			return 0, ErrUserAborted
-		}
 		if model.Choice == "" {
-			return 0, errors.New("no replication execution selected")
+			errChan <- errors.New("user aborted selection")
+			return
 		}
 		// Extract the ID from model.Choice
 		var execID int64
 		_, err = fmt.Sscanf(model.Choice, "ID: %d", &execID)
 		if err != nil {
-			return 0, fmt.Errorf("error parsing execution ID: %w", err)
+			errChan <- fmt.Errorf("error parsing execution ID: %w", err)
+			return
 		}
-		return execID, nil
+		choice <- execID
+		return
 	}
 
-	return 0, errors.New("unexpected program result")
+	errChan <- errors.New("unexpected program result")
 }
--- a/pkg/views/replication/execution/select/view.go
+++ b/pkg/views/replication/execution/select/view.go
@@ -14,8 +14,8 @@
 package execution
 
 import (
+	"errors"
 	"fmt"
-	"os"
 
 	"github.com/charmbracelet/bubbles/list"
 	tea "github.com/charmbracelet/bubbletea"
@@ -23,7 +23,9 @@ import (
 	"github.com/goharbor/harbor-cli/pkg/views/base/selection"
 )
 
-func ReplicationExecutionList(executions []*models.ReplicationExecution, choice chan<- int64) {
+var ErrUserAborted = errors.New("user aborted selection")
+
+func ReplicationExecutionList(executions []*models.ReplicationExecution) (int64, error) {
 	itemsList := make([]list.Item, len(executions))
 	for i, p := range executions {
 		displayName := fmt.Sprintf("ID: %d, Status: %s, Trigger: %s, Start Time: %s, Succeed: %d, Total: %d",
@@ -35,18 +37,24 @@ func ReplicationExecutionList(executions []*models.ReplicationExecution, choice 
 
 	p, err := tea.NewProgram(m, tea.WithAltScreen()).Run()
 	if err != nil {
-		fmt.Println("Error running program:", err)
-		os.Exit(1)
+		return 0, fmt.Errorf("error running selection program: %w", err)
 	}
 
-	if p, ok := p.(selection.Model); ok {
-		// Extract the ID from p.Choice
-		var execID int64
-		_, err = fmt.Sscanf(p.Choice, "ID: %d", &execID)
-		if err != nil {
-			fmt.Println("error parsing execution ID:", err)
-			os.Exit(1)
+	if model, ok := p.(selection.Model); ok {
+		if model.Aborted {
+			return 0, ErrUserAborted
 		}
-		choice <- execID
+		if model.Choice == "" {
+			return 0, errors.New("no replication execution selected")
+		}
+		// Extract the ID from model.Choice
+		var execID int64
+		_, err = fmt.Sscanf(model.Choice, "ID: %d", &execID)
+		if err != nil {
+			return 0, fmt.Errorf("error parsing execution ID: %w", err)
+		}
+		return execID, nil
 	}
+
+	return 0, errors.New("unexpected program result")
 }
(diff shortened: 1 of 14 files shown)


gcharpe1604:
I’ve pushed a larger round of updates after additional verification of the replication and tag retention flows.

The changes now address several related issues around:

* graceful cancellation handling (`ctrl+c` / `esc`)
* error propagation instead of abrupt exits
* panic prevention for failed API fetches
* safer retention rule deletion behavior
* proper support for `--project-id`
* simplification of the replication prompt flow by removing unnecessary goroutine/channel handling

One important fix here was preventing cancelled selection prompts from unintentionally proceeding with default rule indices during tag retention deletion flows.

The updated implementation now consistently bubbles errors/cancellations back through Cobra instead of terminating inside prompt/view helpers.

@qcserestipy @NucleoFusion

Would appreciate your feedback/review whenever you have time. Thanks!

qcserestipy:
@gcharpe1604 Thank you for your contribution. However, this PR addresses much more changes than the connected issue. Please reduce this PR to the mentioned changes in the issue. For other changes open a new issue and related PRs.

gcharpe1604:
@qcserestipy
Thanks for the review and feedback, and apologies for the delay in following up.

I had university end-semester exams over the last week and wasn't able to spend time on the PR. I'm reviewing the feedback now and re-evaluating the scope based on your comments.

I'll reduce the PR so it stays focused on the issue being addressed. The broader prompt/view improvements and related changes can be proposed separately where appropriate to keep the review scope manageable.

gcharpe1604:
Hi @qcserestipy,

I have updated this PR branch to keep it strictly scoped to the replication-related prompt and view error handling changes as suggested.

### Key Changes
1. Replaced direct `os.Exit(1)` and `log.Fatal()` calls inside replication selection views and prompt helpers with standard error propagation.
2. Propagated empty choices (for example when users cancel a selection) back to the caller commands as errors instead of continuing with invalid values.

All unrelated changes (retention fixes, broader refactoring, and base selection model updates) have been removed from this branch.

Could you please take another look when you have a chance?

Thanks!!!

qcserestipy:
@gcharpe1604 Thanks for making those changes, I will have a look. To make this PR more complete please add a screenshot or screen recording of the failing and resolved behavior

gcharpe1604:
Here is the comparison showing the behavior before and after this change:

### 1. Before (`harbor_main.exe`)
When running commands that prompt the user, if there is a connection refusal or API client error, the CLI exits abruptly and silently (exits with code `1` but prints nothing to stderr/stdout because the fatal logging is bypassed by default logrus configurations):
```powershell
PS C:\Users\govin\OneDrive\Documents\opensrc\harbor-cli> .\harbor_main.exe replication start
PS C:\Users\govin\OneDrive\Documents\opensrc\harbor-cli> 
```

https://github.com/user-attachments/assets/2f89dae1-8cbf-48b2-b7d1-8c5465313f02


### 2. After (`harbor_fixed.exe`)
With this change, the prompt-related errors are correctly propagated through Cobra's `RunE` method, printing a proper CLI error message and exiting gracefully:
```powershell
PS C:\Users\govin\OneDrive\Documents\opensrc\harbor-cli> .\harbor_fixed.exe replication start
Error: failed to get replication policy: Get "http://localhost:8080/api/v2.0/replication/policies?page=0&page_size=0": dial tcp [::1]:8080: connectex: No connection could be made because the target machine actively refused it.
```

https://github.com/user-attachments/assets/fb3c5a9e-b05d-472b-8093-9b56c3419a21

gcharpe1604:
@qcserestipy Hi, I’ve added the requested before/after recordings and details. Could you please take another look when you have time? Thanks!

qcserestipy:
Please address the changes requested by Copilot, otherwise lgtm @gcharpe1604

gcharpe1604:
@qcserestipy Thanks for the review! I’ll address the Copilot suggested changes and push an update soon.

gcharpe1604:
Addressed the Copilot review comments in the latest commit:

* reused `ErrUserAborted` for abort paths
* added validation for the replication policy lookup to avoid silently returning ID `0`
* verified with `go build ./...`

@qcserestipy Please review when you get a chance.

pkg/views/replication/policies/select/view.go
```diff
@@ -14,16 +14,18 @@
...
-	"os"
 
 	"github.com/charmbracelet/bubbles/list"
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/goharbor/go-client/pkg/sdk/v2.0/models"
 	"github.com/goharbor/harbor-cli/pkg/views/base/selection"
 )
 
-func ReplicationPoliciesList(policies []*models.ReplicationPolicy, choice chan<- int64) {
+var ErrUserAborted = errors.New("user aborted selection")
```
qcserestipy: `ErrUserAborted` is now declared identically in all three replication select views, and #996 adds a fourth copy in the retention view. could you pull this into one shared sentinel (e.g. in `pkg/views/base/selection`) and reuse it everywhere? otherwise callers cant `errors.Is` against a single value and we keep copy-pasting it.

pkg/views/replication/execution/select/view.go
```diff
@@ -35,18 +37,25 @@ func ReplicationExecutionList(executions []*models.ReplicationExecution, choice
...
-		// Extract the ID from p.Choice
+	if model, ok := p.(selection.Model); ok {
+		if model.Choice == "" {
+			errChan <- ErrUserAborted
+			return
+		}
+		// Extract the ID from model.Choice
 		var execID int64
-		_, err = fmt.Sscanf(p.Choice, "ID: %d", &execID)
+		_, err = fmt.Sscanf(model.Choice, "ID: %d", &execID)
```
qcserestipy: parsing the id back out of the formatted display string with Sscanf is a bit fragile, it breaks if the display format ever changes. the policies view in this same PR uses a `map[display]int64` which is more robust, could you use the same approach here for consistency.

qcserestipy (changes requested):
Thank you for your contribution! this is going in the right direction and the copilot points look addressed. two things before merge:

- the new error/abort paths arent covered by any tests, which is also why codecov/patch is failing. could you add tests for these? the webhook and retention delete PRs use package-level vars as test seams for exactly this, would be good to follow that.
- details on the duplicated sentinel and the Sscanf parsing inline.

gcharpe1604:
Addressed both comments.

I moved `ErrUserAborted` to the shared `pkg/views/base/selection` package and updated the three replication selection views to use it.

I also replaced the `Sscanf` parsing in the execution selector with a checked display-to-ID map, while keeping the current display format unchanged.

`go test ./pkg/views/...` and `go test ./...` both pass.

PR #944, opened 2026-05-18
fix(login): support --password-stdin for piped input

### LINKED ISSUE #943: [bug]: support --password-stdin for piped input and preserve whitespace
## Description
An issue with harbor login --password-stdin when password is provided via piped input.
Previously, the login flow used term.ReadPassword(---), which expects an interactive terminal (TTY). This caused failures or incorrect behavior when stdin was piped (non-interactive).

This change introduces GetSecretStdin(...) which:

Detects whether stdin is interactive or piped
Uses hidden password input for interactive terminals
Reads directly from stdin when input is piped

## Steps to Reproduce
Run command:
- echo "mypassword" | harbor login example.com -u user --password-stdin
Observe behavior

## Expected Behavior
CLI should correctly read password from stdin when piped
Password should not be modified (including leading/trailing spaces)

## Actual Behavior
Login may fail or behave incorrectly when stdin is piped
Password may be altered due to TrimSpace removing valid whitespace

## Environment
- OS: macOS (also reproducible on other platforms)
- Tool version: harbor-cli (main branch before fix)
- Other relevant details: stdin is non-interactive when using pipe

## Additional Context
Note: Running full go test ./pkg/utils may fail in sandbox environments due to attempts to write under $HOME/.harbor, which is unrelated to this change..

rambhardwajj:
## Description
Fixes an issue with harbor login --password-stdin when password is provided via piped input
Previously, the login flow relied on term.ReadPassword(...), which expects an interactive terminal (TTY). This caused failures or incorrect behavior when stdin was non-interactive (e.g., piped input).

This change ensures correct handling of both interactive and piped stdin, and also fixes unintended password modification caused by trimming whitespace.

- Fixes #943 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
- Use GetSecretStdin(...) to handle both interactive and piped stdin input
- Detect non-TTY stdin and read password directly from input stream
- Preserve leading/trailing whitespace in passwords (avoid TrimSpace)
- Trim only newline characters (\r, \n)
- Add unit tests for piped input and whitespace preservation

### DIFF
--- a/cmd/harbor/root/login.go
+++ b/cmd/harbor/root/login.go
@@ -16,7 +16,6 @@ package root
 import (
 	"context"
 	"fmt"
-	"os"
 	"strings"
 
 	"github.com/goharbor/go-client/pkg/harbor"
@@ -28,7 +27,6 @@ import (
 	"github.com/goharbor/harbor-cli/pkg/views/login"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
-	"golang.org/x/term"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


qcserestipy:
Can you please add screenshots or screen recordings that document the current and expected behavior.

qcserestipy:
Since the author did not get back to it

PR #954, opened 2026-05-23
added tag retention tasks list subcommand

NishchayRajput:
## Description
This PR adds the `harbor tag retention tasks list` subcommand to display retention execution tasks for a specific retention policy execution.

It enables users to view and inspect individual tasks from a retention policy execution, showing details such as task ID, repository, status, retention metrics, and timing information.

- Fixes #802

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Refactor
- [ ] Chore / maintenance

## Changes
- Added `harbor tag retention tasks` command group under `harbor tag retention`.
- Added `harbor tag retention tasks list [PROJECT_NAME]` with:
  - Interactive mode (prompts for project and execution selection if not provided)
  - Flag-based mode with explicit project ID/name and execution ID
  - Table-based output with task details (ID, Execution ID, Repository, Status, Retained, Total, Code, Revision, Start Time, End Time)
  - JSON output format support via `--output-format json`
- Added retention API helpers:
  - `ListRetentionExecutions()` - fetch retention executions for a policy
  - `ListRetentionTasks()` - fetch tasks for a specific execution
- Added UI/View components:
  - `RetentionExecutionList()` - interactive selection view for executions
  - `RetentionTaskList()` - task details table view
- Added prompt helpers:
  - `GetRetentionExecutionIDFromUser()` - interactive execution selection

### DIFF
--- a/cmd/harbor/root/tag/retention/tasks/cmd.go
+++ b/cmd/harbor/root/tag/retention/tasks/cmd.go
@@ -1,0 +1,30 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package retention
+
+import "github.com/spf13/cobra"
+
+func TasksCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "tasks",
+		Short: "Manage retention execution tasks",
+		Long:  "List retention execution tasks",
+	}
+
+	cmd.AddCommand(
+		ListCommand(),
+	)
+
+	return cmd
+}
--- a/cmd/harbor/root/tag/retention/tasks/list.go
+++ b/cmd/harbor/root/tag/retention/tasks/list.go
@@ -1,0 +1,95 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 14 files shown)


cmd/harbor/root/tag/retention/tasks/cmd.go
```diff
@@ -0,0 +1,30 @@
...
+// limitations under the License.
+package retention
+
+import "github.com/spf13/cobra"
+
+func TasksCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "tasks",
+		Short: "Manage retention execution tasks",
+		Long:  "List retention execution tasks",
```
NucleoFusion, line 22: this shouldnt mention only list, that should be in the ListCommand()

cmd/harbor/root/tag/retention/tasks/list.go
```diff
@@ -0,0 +1,95 @@
...
+			}
+
+			taskview.ListRetentionTasks(resp.Payload)
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVarP(&projectName, "project-name", "p", "", "Project name")
+	cmd.Flags().IntVarP(&projectID, "project-id", "i", -1, "Project ID")
+	cmd.Flags().Int64VarP(&executionID, "execution-id", "e", -1, "Retention execution ID")
```
NucleoFusion, line 92: I think taking defaults to be 0 (as ints usually are) and then adding an extra validation somewhere that they arent negative might be the play.
So, we can check if its `== 0` instead of `== -1` and then also raise an invalid error if ` < 0 `.

pkg/prompt/prompt.go
```diff
@@ -505,3 +506,25 @@ func GetPreheatPolicyNameFromUser(projectName string) (string, error) {
...
+		}
+
+		length <- len(response.Payload)
+		rexecView.RetentionExecutionList(response.Payload, executionID)
+	}()
+
+	if <-length == 0 {
+		return 0
+	}
+
```
NucleoFusion, line 528: You can just make this not be in a goroutine
Just use normally, since this isnt really async anyway.

NucleoFusion (changes requested):

PR #955, opened 2026-05-23
fix(registry): return error when registry name not found

### LINKED ISSUE #933: GetRegistryIdByName returns (0, nil) for non-existent registry; callers delete or operate on ID 0
## Problem

`GetRegistryIdByName` returns `(0, nil)` when no registry matches the provided name.

```go
return 0, err
```

At this point `err` is already `nil`, so callers interpret the result as success and continue operating on registry ID `0`.

## Impact

Commands like:

```bash
harbor registry delete does-not-exist
```

silently call:

```text
DELETE /registries/0
```

instead of returning a user-facing error.

This creates:

* confusing UX
* silent invalid API requests
* potential wrong-resource operations

## Affected Areas

* `cmd/harbor/root/registry/delete.go`
* `cmd/harbor/root/registry/update.go`
* `cmd/harbor/root/replication/policies/create.go`

## Root Cause

`GetRegistryIdByName` returns:

```go
return 0, err
```

after iterating all registries without finding a match.

Since `err == nil`, callers receive `(0, nil)`.

## Expected Behavior

When a registry name is not found:

```go
return 0, fmt.Errorf("registry with name %q not found", registryName)
```

should be returned.

## Reproduction

```bash
harbor registry delete does-not-exist
```

Expected:

```text
Error: registry with name "does-not-exist" not found
```

Actual:

```text
DELETE /registries/0
```

(or Harbor API 404 response)

## Proposed Fix

1. Return explicit error from `GetRegistryIdByName`
2. Remove ignored errors (`_`) in callers
3. Properly propagate errors to users

sahu-virendra-1908:
## Description

Fixes an issue where `GetRegistryIdByName` returned `(0, nil)` when a registry name did not exist.

Previously, callers interpreted the result as success and continued operating on registry ID `0`, causing commands like:

```bash
harbor registry delete does-not-exist
```

to incorrectly call:

```text
DELETE /registries/0
```

This change now returns a proper error when the registry is not found and propagates the error correctly in the delete command.

* Fixes #933

## Type of Change

* [x] Bug fix
* [ ] New feature
* [ ] Refactor
* [ ] Documentation update
* [ ] Chore / maintenance

## Changes

* Return explicit error when registry name is not found in `GetRegistryIdByName`
* Handle and propagate errors correctly in `registry/delete.go`
* Prevent invalid operations on registry ID `0`

### DIFF
--- a/cmd/harbor/root/registry/delete.go
+++ b/cmd/harbor/root/registry/delete.go
@@ -33,7 +33,10 @@ func DeleteRegistryCommand() *cobra.Command {
 			errChan := make(chan error, len(args))
 			if len(args) > 0 {
 				for _, arg := range args {
-					registryID, _ := api.GetRegistryIdByName(arg)
+					registryID, err := api.GetRegistryIdByName(arg)
+					if err != nil {
+						return err
+					}
 					wg.Add(1)
 					go func(registryID int64) {
 						defer wg.Done()
--- a/pkg/api/registry_handler.go
+++ b/pkg/api/registry_handler.go
@@ -188,5 +188,8 @@ func GetRegistryIdByName(registryName string) (int64, error) {
 		}
 	}
 
-	return 0, err
+	return 0, fmt.Errorf(
+    "registry with name %q not found",
+    registryName,
+)
 }


sahu-virendra-1908:
@bupd @Vad1mo @OrlinVasilev Please review this PR when you have time. I’ve addressed the issue by returning a proper error when the registry name is not found and by propagating the error correctly in the delete flow.

If any additional changes or improvements are needed, please let me know. Thanks

cmd/harbor/root/registry/delete.go
```diff
@@ -33,7 +33,10 @@ func DeleteRegistryCommand() *cobra.Command {
 			errChan := make(chan error, len(args))
 			if len(args) > 0 {
 				for _, arg := range args {
-					registryID, _ := api.GetRegistryIdByName(arg)
+					registryID, err := api.GetRegistryIdByName(arg)
+					if err != nil {
+						return err
```
qcserestipy, line 38: please return a formated err that also parses the upstream harbor error message. There is `utils.ParseHarborErrorMsg` that you can use

pkg/api/registry_handler.go
```diff
@@ -187,5 +188,8 @@ func GetRegistryIdByName(registryName string) (int64, error) {
 		}
 	}
 
-	return 0, err
+	return 0, fmt.Errorf(
+    "registry with name %q not found",
```
qcserestipy, line 192: please fix the format of these new lines

also use `utils.ParseHarborErrorMsg`

qcserestipy (changes requested):
Thank you for your contribution! In principle LGTM, please fix the small suggestions

qcserestipy:
Please also fix your failing DCO check

qcserestipy:
Since there was no feedback

PR #956, opened 2026-05-23
fix: key checksums by filename to include windows_amd64

### LINKED ISSUE #950: [bug]: Checksum file doesnt include 'windows-amd64' in release 0.20 & in-turn breaks scoop
## Description

Scoop relies on getting the checksums from checksum.txt for updating its manifest. But right now, in the checksum there is no entry for amd64.zip Which breaks scoop.

## Steps to Reproduce
1. Download Checksum file from Releases 0.0.20
2. Open and Verify amd64_windows

cotishq:
## Description

Scoop autoupdate reads hashes from `checksums.txt` by artifact filename. On v0.0.20, `harbor-cli_*_windows_amd64.zip` was missing from that file because duplicate SHA256 values overwrote each other when checksums were keyed by hash.

- Fixes #950 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
-  Key checksum entries by filename in `.dagger/checksum.go` instead of by hash
-  Emit `checksums.txt` with filenames sorted alphabetically
- Both `windows_amd64` and `windows_arm64` archives now appear in `checksums.txt` when their hashes collide

## Test plan
-  [x] `dagger call build` -> `archive` -> `nfpm-build` -> `apk` -> `checksum`
---
-  [x] `grep windows dist/checksums.txt` shows both `windows_amd64.zip` and `windows_arm64.zip`
Screenshot From 2026-05-24 00-12-51

---
-  [x] `sha256sum` on `dist/archive/*_windows_*.zip` matches `checksums.txt`
image

### DIFF
--- a/.dagger/checksum.go
+++ b/.dagger/checksum.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 
 	"dagger/harbor-cli/internal/dagger"
@@ -36,12 +37,19 @@ func (m *HarborCli) Checksum(ctx context.Context,
 		}
 
 		split := strings.Split(v, "/")
-		sums[out] = split[len(split)-1] // Taking only filename
+		filename := split[len(split)-1]
+		sums[filename] = strings.TrimSpace(out)
 	}
 
+	filenames := make([]string, 0, len(sums))
+	for filename := range sums {
+		filenames = append(filenames, filename)
+	}
+	sort.Strings(filenames)
+
 	content := ""
-	for k, v := range sums {
-		content += fmt.Sprintf("%s %s\n", strings.TrimSuffix(k, "\n"), v)
+	for _, filename := range filenames {
+		content += fmt.Sprintf("%s %s\n", sums[filename], filename)
 	}
 
 	buildDir = buildDir.WithFile("checksums.txt", dag.File("checksums.txt", content))


cotishq:
@NucleoFusion would appreciate your review on this

bhavyaKhatri2703:
But why we got duplicate hash values in the first place tho?

NucleoFusion:
I will review this in a while but I'm curious t know what the issue was

cotishq:
> But why we got duplicate hash values in the first place tho?

so the duplicate hash occurs because the two Windows archives (windows_amd64.zip and windows_arm64.zip) can produce identical SHA256 values when they contain the same binary content - for example if the cross-compiled binaries for both targets are functionally identical at that release. The original code used the hash as the map key, so the second entry silently overwrote the first. Keying by filename instead ensures both entries are always written regardless of hash collision.

cotishq:
> I will review this in a while but I'm curious t know what the issue was

the issue was like the checksums were keyed by hash value in a map, so when both the zips( amd64 and arm64 ) produced the same sha256, the second entry overwrote the first before being written to checksums.txt. 
so i just fixed by keying the map by filename instead, so now both entries always appear regardless of hash collision

bhavyaKhatri2703:
> > But why we got duplicate hash values in the first place tho?
> 
> so the duplicate hash occurs because the two Windows archives (windows_amd64.zip and windows_arm64.zip) can produce identical SHA256 values when they contain the same binary content - for example if the cross-compiled binaries for both targets are functionally identical at that release. The original code used the hash as the map key, so the second entry silently overwrote the first. Keying by filename instead ensures both entries are always written regardless of hash collision.

yeah that makes sense

NucleoFusion:
I see, that was kinda stupid on my part then lol
Is the sha's for both still the same? Or are they different now?
Do you have a sample checksum file you can show?

cotishq:
yeah no worries, the sha's are still the same , but i think thats expected and not a problem
scoop just needs both filenames present in the file regardless of whether the hashes match
here is the ss of the sample output of checksums.txt 
image

cotishq:
hey @NucleoFusion , can you tell me like is it a valid fix or i have to still iterate it ?

NucleoFusion:
idk why the shasum is same for both tho, it should be different
maybe its connected to the #970 maybe we should wait for that

kkrypt0nn:
Since the binaries are missing in the archive, both return the same hash as both contain only and exactly the same `README.md` and `LICENSE.md` file

Should be fixed with #970 as cross compiled binaries do not return the same hash, it's compiled for a whole different CPU architecture

bupd:
Thanks for your contribution, closing this since issue root cause is already fixed with #970

PR #965, opened 2026-05-26
fix(ci): retry cosign transparency log failures

### LINKED ISSUE #964: [bug]: image signing fails on transient Sigstore transparency log errors
> *This was generated by AI during triage.*

## Description
The main branch image publish job intermittently fails after the image has already been pushed because `cosign attest` exits on Sigstore/Rekor transparency-log errors.

## Motivation
Release and main-branch image publishing should be safe to rerun. A transient Sigstore transport failure, or an already-created transparency-log entry from a previous attempt, should not leave the whole pipeline red after the image publish has succeeded.

## Root Cause
`.dagger/publishimage.go` runs `cosign attest` and `cosign sign` exactly once with `--timeout 1m`. The current failing main run published `registry.goharbor.io/harbor-cli/harbor-cli:latest@sha256:776de248b4fbcff8ffe7603771c877050a46313b0c3407df57cb05c549091f89`, then failed during `cosign attest` with:

```text
signing bundle: stream error: stream ID 1; INTERNAL_ERROR; received from peer
```

A prior release rerun hit the same code path with Rekor `409 createLogEntryConflict`, which means the transparency-log entry already existed but the pipeline still failed.

## Expected Behavior
Image signing and SBOM attestation should retry transient Sigstore/Rekor failures and treat duplicate transparency-log entries as successful completion.

## Actual Behavior
Any single `cosign attest` or `cosign sign` failure exits the Dagger function and fails the workflow.

## References
- Failing main run: https://github.com/goharbor/harbor-cli/actions/runs/26455099188

bupd:
## Summary
- Retry `cosign attest` and `cosign sign` when Sigstore/Rekor returns transient transport errors such as stream `INTERNAL_ERROR`.
- Treat Rekor duplicate transparency-log entries as success so reruns do not fail after an entry was already created.
- Increase the per-cosign timeout from 1 minute to 2 minutes for signing bundle creation.

## Root Cause
The image publish step succeeds before the workflow enters Sigstore/Rekor signing. The Dagger module ran each `cosign` command once, so transient transparency-log errors or duplicate entries from reruns failed the whole job even when the publish side had completed.

## Testing
- `git diff --check`
- `dagger call publish-image-and-sign --help`

`go test ./...` under `.dagger` is not runnable from a plain checkout because the generated `internal/dagger` package is gitignored and absent until Dagger loads the module.

Closes #964

### DIFF
--- a/.dagger/publishimage.go
+++ b/.dagger/publishimage.go
@@ -43,10 +43,10 @@ func (m *HarborCli) PublishImageAndSign(
 		// Generate SBOM (SPDX JSON) for the published image
 		sbom := m.GenerateSBOM(
 			ctx,
-			addr, 
-			registryUsername, 
+			addr,
+			registryUsername,
 			registryPassword,
-		);
+		)
 
 		// Attest SBOM to the image using Cosign in-toto attestation
 		if _, err := m.AttestSBOM(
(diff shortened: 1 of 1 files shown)


bupd:
@codex review

PR #996, opened 2026-06-14
fix(retention): prevent failures in tag retention rule deletion flow

### LINKED ISSUE #995: [bug]: Errors not propagated in tag retention rule deletion flow
### Description
The `tag retention delete` command lacks error handling when retrieving and selecting the tag retention rules. Additionally, `api.DeleteRetention()` performs redundant network queries to resolve project names to retention IDs.

### Environment
- Version: Harbor CLI latest main branch
- OS: Windows / macOS / Linux

### Steps To Reproduce
1. Log in to a dummy Harbor server address (which fails connection) using `harbor login --skip-verify-client`.
2. Run the command: `harbor tag retention delete --project <project_name>`.
3. Select the retention rule from the prompt or observe the behavior when the API server is unreachable.

### Expected Behavior
- The CLI should surface a clear network connection error or selection cancellation error to the user and exit with code `1`.
- The API helper should utilize direct retention IDs when already available to save network lookups.

### Actual Behavior
- The CLI exits unexpectedly, blocks, or fails to surface the underlying network/connection error to the user when the rule retrieval fails.
- The `DeleteRetention` client helper makes a redundant network call to `GetRetentionId` even when the retention ID is already resolved by the caller.

### Evidence
```powershell
PS C:\Users\govin\OneDrive\Documents\opensrc\harbor-cli> .\harbor.exe tag retention delete --project test-project
# Silent exit (exits with code 1, but prints no errors to stdout or stderr)
```

gcharpe1604:
- Fixes #995 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

---

## Description / Quick Summary
This PR safeguards the tag retention rule deletion flow against unhandled errors. API failures and prompt cancellations are propagated through Cobra's RunE boundary. When no retention rules are configured, the command prints "No retention rules found" and exits successfully without attempting deletion.

---

## Problem
* **What issue exists?** 
  - `prompt.GetRetentionTagRule()` would block indefinitely on connection failures and did not return errors to its caller.
  - The deletion command assumed retention rules were always found, resulting in potential errors when indexing empty payloads.
  - `api.DeleteRetention()` executed redundant network lookups to resolve project names to retention IDs when the ID was already available.
* **Who is affected?** Users attempting to view or delete tag retention rules when network/connection problems occur, or when no retention rules are configured on the project.
* **Current Behavior:** The CLI could exit unexpectedly, block, or fail to surface actionable errors to the user.
* **Desired Behavior:** API failures and prompt cancellations should propagate gracefully up to Cobra with user-friendly errors. Empty retention-rule lists are treated as a successful no-op and print "No retention rules found".

---

## Root Cause
* **Relevant Files:**
  - `cmd/harbor/root/tag/retention/delete.go`
  - `pkg/api/retention_handler.go`
  - `pkg/prompt/prompt.go`
  - `pkg/views/retention/select/view.go`
* **Technical Explanation:**
  The command and API wrapper layers did not check return errors from `api.ListRetention()` and did not propagate the error up to the Cobra `RunE` boundary.

---

## Solution
* **Overview:** 
  1. Updated `prompt.GetRetentionTagRule` signature to return `(int, error)` and check for empty rules.
  2. Simplified `DeleteRetention` to use the already-available retention ID and removed a redundant lookup request.
  3. Propagated selection aborts and API errors up to `cmd/harbor/root/tag/retention/delete.go`.

---

### Included
- Returning errors from `GetRetentionTagRule` and propagating them in the `delete` command.
- Optimizing `DeleteRetention` client helper to use direct ID.

### Explicitly Not Included
- Core selection model modification (no changes are made to `pkg/views/base/selection/model.go` to keep this PR strictly isolated).

---

## Changes
- **`cmd/harbor/root/tag/retention/delete.go`**: Added error check for `prompt.GetRetentionTagRule` and updated `api.DeleteRetention` invocation to use the direct retention ID.
- **`pkg/api/retention_handler.go`**: Simplified `DeleteRetention` signature to accept `retentionID` directly and removed the redundant call to `GetRetentionId`.
- **`pkg/prompt/prompt.go`**: Refactored `GetRetentionTagRule` to return `(int, error)` and added checks for API errors and empty rule lists.
- **`pkg/views/retention/select/view.go`**: Refactored `RetentionList` to return `(int, error)` and propagate selection errors.

---

### Unit Tests
- [x] All existing package tests pass (`go test ./...`)

### Manual Verification
1. Configured local dummy server.
2. Ran `.\harbor_retention.exe tag retention delete --project <project_name>`.
3. Observed the graceful exit and correct error reporting on connection refusal.

---

## Risks & Compatibility
* **Backward Compatibility:** Non-breaking. Enhances error propagation without changing successful prompt behavior.
* **Potential Risks:** None identified.

### DIFF
--- a/cmd/harbor/root/tag/retention/delete.go
+++ b/cmd/harbor/root/tag/retention/delete.go
@@ -75,8 +75,11 @@ Examples:
 					return fmt.Errorf("error retrieving retention policy ID: %w", err)
 				}
 			}
-			retentionIndex := prompt.GetRetentionTagRule(retentionID)
-			err = api.DeleteRetention(projectName, int(retentionIndex))
+			retentionIndex, err := prompt.GetRetentionTagRule(retentionID)
+			if err != nil {
+				return err
+			}
+			err = api.DeleteRetention(retentionID, retentionIndex)
 			if err != nil {
 				return fmt.Errorf("%w", err)
 			}
--- a/cmd/harbor/root/tag/retention/delete.go
+++ b/cmd/harbor/root/tag/retention/delete.go
@@ -80,9 +80,10 @@ Examples:
 				return err
 			}
 			if retentionIndex < 0 {
+				fmt.Println("No retention rules found")
 				return nil
 			}
-			err = api.DeleteRetention(retentionID, retentionIndex)
+			err = api.DeleteRetention(retentionID, int(retentionIndex))
 			if err != nil {
 				return fmt.Errorf("%w", err)
 			}
--- a/cmd/harbor/root/tag/retention/delete.go
+++ b/cmd/harbor/root/tag/retention/delete.go
@@ -79,6 +79,9 @@ Examples:
 			if err != nil {
 				return err
 			}
+			if retentionIndex < 0 {
+				return nil
+			}
 			err = api.DeleteRetention(retentionID, retentionIndex)
 			if err != nil {
 				return fmt.Errorf("%w", err)
--- a/cmd/harbor/root/tag/retention/delete.go
+++ b/cmd/harbor/root/tag/retention/delete.go
@@ -80,15 +80,15 @@ Examples:
 					return fmt.Errorf("error retrieving retention policy ID: %w", err)
 				}
 			}
-			retentionIndex, err := getRetentionTagRule(retentionID)
+			retentionPolicy, retentionRuleID, err := getRetentionTagRule(retentionID)
 			if err != nil {
 				return err
 			}
-			if retentionIndex < 0 {
+			if retentionRuleID < 0 {
 				fmt.Println("No retention rules found")
 				return nil
 			}
-			err = deleteRetention(retentionID, int(retentionIndex))
+			err = deleteRetention(retentionPolicy, retentionRuleID)
 			if err != nil {
 				return fmt.Errorf("%w", err)
 			}
--- a/cmd/harbor/root/tag/retention/delete.go
+++ b/cmd/harbor/root/tag/retention/delete.go
@@ -22,6 +22,11 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var getProjectNameFromUser = prompt.GetProjectNameFromUser
+var getRetentionId = api.GetRetentionId
+var getRetentionTagRule = prompt.GetRetentionTagRule
+var deleteRetention = api.DeleteRetention
+
 func DeleteRetentionRuleCommand() *cobra.Command {
 	var projectName string
 	var projectID int
@@ -49,7 +54,7 @@ Examples:
 			}
 
 			if projectID == -1 && projectName == "" {
-				name, err := prompt.GetProjectNameFromUser()
+				name, err := getProjectNameFromUser()
 				if err != nil {
 					return err
 				}
@@ -66,7 +71,7 @@ Examples:
 				projectIDStr = projectName
 			}
 
-			retentionID, err := api.GetRetentionId(projectIDStr, isName)
+			retentionID, err := getRetentionId(projectIDStr, isName)
 			if err != nil {
 				if err.Error() == "No retention policy exists for this project" {
 					fmt.Println("No retention policy exists for this project")
@@ -75,7 +80,7 @@ Examples:
 					return fmt.Errorf("error retrieving retention policy ID: %w", err)
 				}
 			}
-			retentionIndex, err := prompt.GetRetentionTagRule(retentionID)
+			retentionIndex, err := getRetentionTagRule(retentionID)
 			if err != nil {
 				return err
 			}
@@ -83,7 +88,7 @@ Examples:
 				fmt.Println("No retention rules found")
 				return nil
 			}
-			err = api.DeleteRetention(retentionID, int(retentionIndex))
+			err = deleteRetention(retentionID, int(retentionIndex))
 			if err != nil {
 				return fmt.Errorf("%w", err)
 			}
--- a/pkg/api/retention_handler.go
+++ b/pkg/api/retention_handler.go
@@ -121,18 +121,13 @@ func GetRetentionId(projectNameorID string, isName bool) (string, error) {
 	return retentionid, nil
 }
 
-func DeleteRetention(projectName string, ruleIndex int) error {
+func DeleteRetention(retentionID string, ruleIndex int) error {
 	ctx, client, err := utils.ContextWithClient()
 	if err != nil {
 		return err
 	}
 
-	retentionIDStr, err := GetRetentionId(projectName, true)
-	if err != nil {
-		return err
-	}
-
-	retentionResp, err := ListRetention(retentionIDStr)
+	retentionResp, err := ListRetention(retentionID)
 	if err != nil {
 		return err
 	}
--- a/pkg/api/retention_handler.go
+++ b/pkg/api/retention_handler.go
@@ -121,7 +121,7 @@ func GetRetentionId(projectNameorID string, isName bool) (string, error) {
 	return retentionid, nil
 }
 
-func DeleteRetention(retentionID string, ruleIndex int64) error {
+func DeleteRetention(retentionID string, ruleIndex int) error {
 	ctx, client, err := utils.ContextWithClient()
 	if err != nil {
 		return err
@@ -137,12 +137,11 @@ func DeleteRetention(retentionID string, ruleIndex int64) error {
 		return fmt.Errorf("retention policy is empty")
 	}
 
-	idx := int(ruleIndex)
-	if idx < 0 || idx >= len(existingPolicy.Rules) {
+	if ruleIndex < 0 || ruleIndex >= len(existingPolicy.Rules) {
 		return fmt.Errorf("invalid rule index")
 	}
 
-	copy(existingPolicy.Rules[idx:], existingPolicy.Rules[idx+1:])
+	copy(existingPolicy.Rules[ruleIndex:], existingPolicy.Rules[ruleIndex+1:])
 	existingPolicy.Rules[len(existingPolicy.Rules)-1] = nil
 	existingPolicy.Rules = existingPolicy.Rules[:len(existingPolicy.Rules)-1]
 
--- a/pkg/api/retention_handler.go
+++ b/pkg/api/retention_handler.go
@@ -133,12 +133,18 @@ func DeleteRetention(retentionID string, ruleIndex int64) error {
 	}
 
 	existingPolicy := retentionResp.Payload
+	if existingPolicy == nil {
+		return fmt.Errorf("retention policy is empty")
+	}
+
 	idx := int(ruleIndex)
 	if idx < 0 || idx >= len(existingPolicy.Rules) {
 		return fmt.Errorf("invalid rule index")
 	}
 
-	existingPolicy.Rules = append(existingPolicy.Rules[:idx], existingPolicy.Rules[idx+1:]...)
+	copy(existingPolicy.Rules[idx:], existingPolicy.Rules[idx+1:])
+	existingPolicy.Rules[len(existingPolicy.Rules)-1] = nil
+	existingPolicy.Rules = existingPolicy.Rules[:len(existingPolicy.Rules)-1]
 
 	_, err = client.Retention.UpdateRetention(ctx, &retention.UpdateRetentionParams{
 		ID:     int64(retentionResp.Payload.ID),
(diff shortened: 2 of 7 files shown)


pkg/api/retention_handler.go
```diff
@@ -121,18 +121,13 @@ func GetRetentionId(projectNameorID string, isName bool) (string, error) {
...
 	if err != nil {
 		return err
 	}
 
-	retentionIDStr, err := GetRetentionId(projectName, true)
-	if err != nil {
-		return err
-	}
-
-	retentionResp, err := ListRetention(retentionIDStr)
```
NucleoFusion, line 135: wait, why are we removing this?

pkg/prompt/prompt.go
```diff
@@ -462,13 +462,15 @@ func GetRoleIDFromUser() int64 {
...
-	return <-retentionIndex
+func GetRetentionTagRule(retentionID string) (int, error) {
+	response, err := api.ListRetention(retentionID)
+	if err != nil {
+		return 0, err
+	}
+	if response.Payload == nil || len(response.Payload.Rules) == 0 {
+		return 0, fmt.Errorf("no retention rules found")
+	}
+	return retview.RetentionList(response.Payload.Rules)
```
NucleoFusion: I dont think no retention rules should be an error
We should print "No retention rules found" or something instead of showing `ERROR: No retention rules`
This might be okay since we are just printing as of now, but as the slog refactor / errors package comes through it will get weirder

pkg/views/retention/select/view.go
```diff
@@ -24,9 +24,11 @@ import (
 	"github.com/goharbor/harbor-cli/pkg/views/base/selection"
 )
 
-func RetentionList(rules []*models.RetentionRule, choice chan<- int64) {
+var ErrUserAborted = errors.New("user aborted selection")
+
+func RetentionList(rules []*models.RetentionRule) (int, error) {
 	itemsList := make([]list.Item, len(rules))
-	items := map[string]int64{}
+	items := map[string]int{}
```
NucleoFusion: I think we should continue using int64

NucleoFusion (changes requested):

pkg/api/retention_handler.go
gcharpe1604: The caller in `delete.go` already resolves the retention ID via `GetRetentionId()` before invoking `DeleteRetention`. The removed lines were performing the same lookup a second time. By accepting `retentionID` directly, we eliminate a redundant network call.

pkg/prompt/prompt.go
gcharpe1604: Makes sense, changed it to just print "No retention rules found" and exit cleanly instead of returning an error.

pkg/views/retention/select/view.go
gcharpe1604: Updated to int64, thanks.

pkg/views/retention/select/view.go
```diff
@@ -55,21 +57,28 @@ func RetentionList(rules []*models.RetentionRule, choice chan<- int64) {
 			rule.Template,
 		)
 
-		items[display] = rule.ID
+		items[display] = int64(i)
 		itemsList[i] = selection.Item(display)
 	}
 
 	m := selection.NewModel(itemsList, "Select a Retention Rule")
```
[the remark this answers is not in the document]
gcharpe1604: Addressed in `5b51cf4`: the construct is now passed as `Retention Rule`, avoiding the duplicated prefix.

pkg/views/retention/select/view.go
```diff
@@ -55,21 +57,28 @@ func RetentionList(rules []*models.RetentionRule, choice chan<- int64) {
...
 
-	if p, ok := p.(selection.Model); ok {
-		choice <- items[p.Choice]
+	if model, ok := p.(selection.Model); ok {
+		if model.Aborted {
+			return 0, ErrUserAborted
+		}
+		if model.Choice == "" {
+			return 0, errors.New("no retention rule selected")
+		}
```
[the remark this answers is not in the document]
gcharpe1604: Addressed in `5b51cf4`: an empty choice now returns `ErrUserAborted`.

pkg/prompt/prompt.go
```diff
@@ -462,13 +462,16 @@ func GetRoleIDFromUser() int64 {
...
-	return <-retentionIndex
+func GetRetentionTagRule(retentionID string) (int64, error) {
+	response, err := api.ListRetention(retentionID)
+	if err != nil {
+		return 0, err
+	}
+	if response.Payload == nil || len(response.Payload.Rules) == 0 {
+		fmt.Println("No retention rules found")
+		return -1, nil
+	}
```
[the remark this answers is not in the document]
gcharpe1604: Keeping this as a successful no-op to match the maintainer feedback in the existing review thread: an empty rule list is an expected state, not a command failure.

pkg/api/retention_handler.go
```diff
@@ -121,28 +121,24 @@ func GetRetentionId(projectNameorID string, isName bool) (string, error) {
...
 	if err != nil {
 		return err
 	}
 
 	existingPolicy := retentionResp.Payload
-	if ruleIndex < 0 || ruleIndex >= len(existingPolicy.Rules) {
+	idx := int(ruleIndex)
+	if idx < 0 || idx >= len(existingPolicy.Rules) {
 		return fmt.Errorf("invalid rule index")
 	}
```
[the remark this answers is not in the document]
gcharpe1604: Leaving this unchanged. The value is an internal zero-based slice index selected from the current rules and is bounds-checked immediately before use; the UI displayed one-based ordinal is not passed here.

pkg/api/retention_handler.go
NucleoFusion: okay, nice

pkg/api/retention_handler.go
```diff
@@ -121,7 +121,7 @@ func GetRetentionId(projectNameorID string, isName bool) (string, error) {
 	return retentionid, nil
 }
 
-func DeleteRetention(retentionID string, ruleIndex int) error {
+func DeleteRetention(retentionID string, ruleIndex int64) error {
```
NucleoFusion: I did not mean the index to be int64, this has added unneeded complexity

pkg/prompt/prompt.go
```diff
@@ -462,13 +462,16 @@ func GetRoleIDFromUser() int64 {
...
-		retview.RetentionList(response.Payload.Rules, retentionIndex)
-	}()
-	return <-retentionIndex
+func GetRetentionTagRule(retentionID string) (int64, error) {
+	response, err := api.ListRetention(retentionID)
+	if err != nil {
+		return 0, err
+	}
+	if response.Payload == nil || len(response.Payload.Rules) == 0 {
+		fmt.Println("No retention rules found")
```
NucleoFusion: I did not mean print here, This is for API Call only.
print should be done in the root command.

NucleoFusion (changes requested):

pkg/api/retention_handler.go
gcharpe1604: Fixed, reverted back to `int` and cleaned up the `idx` intermediate variable too since it's no longer needed.

pkg/prompt/prompt.go
gcharpe1604: Fixed, moved the print to `delete.go`. The prompt helper now just returns `(-1, nil)` on empty rules and the command layer handles the user-facing message.

gcharpe1604:
Addressed the latest Copilot review comments:

- removed the dead "model.Aborted" check
- added retention ID context to the empty policy payload error
- wrapped retention-rule listing errors with operation-specific context
- added test seams and RunE coverage for retention deletion paths
- updated the PR description to clarify that empty rule lists are treated as a successful no-op, matching earlier maintainer feedback

CI is passing. Please review when you get a chance.

cmd/harbor/root/tag/retention/delete.go
```diff
@@ -75,8 +75,15 @@ Examples:
...
-			retentionIndex := prompt.GetRetentionTagRule(retentionID)
-			err = api.DeleteRetention(projectName, int(retentionIndex))
+			retentionIndex, err := prompt.GetRetentionTagRule(retentionID)
+			if err != nil {
+				return err
+			}
+			if retentionIndex < 0 {
+				fmt.Println("No retention rules found")
+				return nil
+			}
```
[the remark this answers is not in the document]
gcharpe1604: The successful no-op is intentional and follows the earlier maintainer feedback that an empty retention-rule list is an expected state, not a command failure. I’ve clarified the PR description: the command prints "No retention rules found" and exits with code 0 without attempting deletion.

pkg/prompt/prompt.go
```diff
@@ -462,13 +462,15 @@ func GetRoleIDFromUser() int64 {
...
-	go func() {
-		response, _ := api.ListRetention(retentionID)
-		retview.RetentionList(response.Payload.Rules, retentionIndex)
-	}()
-	return <-retentionIndex
+func GetRetentionTagRule(retentionID string) (int64, error) {
+	response, err := api.ListRetention(retentionID)
+	if err != nil {
+		return 0, err
+	}
```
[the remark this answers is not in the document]
gcharpe1604: The successful no-op is intentional and follows the earlier maintainer feedback that an empty retention-rule list is an expected state, not a command failure. I’ve clarified the PR description: the command prints "No retention rules found" and exits with code 0 without attempting deletion.

pkg/api/retention_handler.go
```diff
@@ -121,28 +121,29 @@ func GetRetentionId(projectNameorID string, isName bool) (string, error) {
...
 		return err
 	}
 
-	retentionResp, err := ListRetention(retentionIDStr)
-	if err != nil {
-		return err
+	existingPolicy := retentionResp.Payload
+	if existingPolicy == nil {
+		return fmt.Errorf("retention policy is empty")
 	}
```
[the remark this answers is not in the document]
gcharpe1604: The successful no-op is intentional and follows the earlier maintainer feedback that an empty retention-rule list is an expected state, not a command failure. I’ve clarified the PR description: the command prints "No retention rules found" and exits with code 0 without attempting deletion.

pkg/views/retention/select/view.go
```diff
@@ -55,21 +57,25 @@ func RetentionList(rules []*models.RetentionRule, choice chan<- int64) {
 			rule.Template,
 		)
 
-		items[display] = rule.ID
+		items[display] = int64(i)
```
qcserestipy: switching from `rule.ID` to the loop index `i` here is risky. the index then gets used in `DeleteRetention` after a *second* `ListRetention` call, so if the rule order differs between the two fetches (concurrent edit, or the api not guaranteeing stable order) you delete the wrong rule. since this is a destructive op id rather keep keying on `rule.ID` and have `DeleteRetention` remove the rule matching that id, so it doesnt depend on ordering.

pkg/api/retention_handler.go
```diff
@@ -121,28 +121,29 @@ func GetRetentionId(projectNameorID string, isName bool) (string, error) {
...
 
-func DeleteRetention(projectName string, ruleIndex int) error {
+func DeleteRetention(retentionID string, ruleIndex int) error {
 	ctx, client, err := utils.ContextWithClient()
 	if err != nil {
 		return err
 	}
 
-	retentionIDStr, err := GetRetentionId(projectName, true)
+	retentionResp, err := ListRetention(retentionID)
```
qcserestipy: this re-lists the retention rules that the prompt already fetched. if you pass the payload (or the selected rule id) down from `GetRetentionTagRule` instead, you drop this extra call and also fix the positional-index issue in select/view.go in one go.

qcserestipy (changes requested):
Thank you for your contribution! this is close, most of the earlier points are handled. one thing left though, the switch to a positional index for deletion can delete the wrong rule if the rule order changes between the two list calls. commented inline, keying on rule.ID would be safer for a destructive op.

gcharpe1604:
Addressed both comments.

The selector now returns the stable retention rule ID instead of the list index. The already-fetched retention policy is passed through to `DeleteRetention`, so it no longer makes a second `ListRetention` call.

`DeleteRetention` now finds and removes the rule by ID, handles nil entries, and returns an error if the selected rule is no longer present.

I also added tests for reordered rules, missing rule IDs, nil entries, and the empty-rule no-op path. `go test ./...` passes.

PR #998, opened 2026-06-15
fix(robot): handle empty permissions in robot view

### LINKED ISSUE #997: [bug]: robot view panics when robot has empty permissions
## Description

`harbor robot view` panics when the Harbor API returns a robot with an empty or nil `permissions` array. The view indexes `robot.Permissions[0]` unconditionally, but the generated `Robot` model treats `permissions` as optional.

`harbor robot list` already guards this case, but the view command does not.

## Steps to Reproduce

1. Run `harbor robot view <robot-id>` against a Harbor instance that returns a robot with `permissions: []`, **or**
2. At code level, call the robot view with `&models.Robot{Permissions: nil}`.

## Expected Behavior

The command should render the robot summary and permissions table, using placeholders (e.g. `0` permissions, `--` namespace) when permission details are missing.

## Actual Behavior

The CLI panics with an index-out-of-range error while rendering the view.

## Environment

- OS: All
- Tool version: current `main`
- Affected code: `pkg/views/robot/view/view.go` (lines ~90 and ~130)

## Additional Context

`pkg/views/robot/list/view.go` already handles empty permissions:

```go
var TotalPermissions int = 0
if len(robot.Permissions) > 0 {
    TotalPermissions = len(robot.Permissions[0].Access)
}
```

The view should follow the same pattern.

**Suggested fix:**
1. Guard `robot.Permissions[0]` access in the summary row and project namespace header.
2. Add a unit test covering nil/empty permissions.

Jay2006sawant:
## Description

- Fixes #997

`harbor robot view` panics when the Harbor API returns a robot with an empty or nil `permissions` array. The view indexed `robot.Permissions[0]` unconditionally, while `harbor robot list` already guarded this case.

This PR adds safe fallbacks for the permission count and project namespace header, matching the list view behavior.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes

- Add `robotPermissionSummary` helper to safely read permission count and namespace
- Use placeholders (`0`, `--`) when permission details are missing
- Add unit tests for nil/empty and populated permissions

## Test plan

- [x] `go test ./pkg/views/robot/view/... -v`
- [x] Verified `robotPermissionSummary` returns safe defaults for nil/empty permissions (no index panic)

### DIFF
--- a/pkg/views/robot/view/view.go
+++ b/pkg/views/robot/view/view.go
@@ -87,7 +87,8 @@ func ViewRobot(robot *models.Robot) {
 		enabledStatus = views.GreenStyle.Render("Enabled")
 	}
 
-	TotalPermissions := strconv.FormatInt(int64(len(robot.Permissions[0].Access)), 10)
+	totalPermissions, namespace := robotPermissionSummary(robot)
+	TotalPermissions := strconv.FormatInt(int64(totalPermissions), 10)
 
 	if robot.ExpiresAt == -1 {
 		expires = "Never"
@@ -127,7 +128,7 @@ func ViewRobot(robot *models.Robot) {
 		permissionsColumns = projectPermissionsColumns
 		resourceStrings = projectResourceStrings
 		systemLevel = false
-		fmt.Printf("\n%s\n\n", views.BoldStyle.Render(fmt.Sprintf("System-level robot with access across projects: %s", robot.Permissions[0].Namespace)))
+		fmt.Printf("\n%s\n\n", views.BoldStyle.Render(fmt.Sprintf("System-level robot with access across projects: %s", namespace)))
 	}
 
 	var permissionRows []table.Row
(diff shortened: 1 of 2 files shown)


Jay2006sawant:
@Vad1mo Can you please look into this?

NucleoFusion (commented):
I dont think so we can create robots with 0 permissions. (I confirmed you cant)
But i think the gaurd is not a bad to have.
@qcserestipy what do you think? Should we have it or is it redundant?

qcserestipy:
@NucleoFusion I think we could get this into the code. I would be curious to know how this issue was found if the cli and ui both do not allow to create robots with 0 permissions in practice @Jay2006sawant.

qcserestipy (approved):
Thank you for your contribution! i think the guard is worth keeping. `robot list` already guards this the same way (`pkg/views/robot/list/view.go`), so this just brings the view in line with it. and even if the ui doesnt let you create a robot with 0 permissions, this is api response data, so a view shouldnt index Permissions[0] unconditionally anyway. cheap check, avoids a panic. lgtm.

bupd (approved):
/lgtm

PR #1002, opened 2026-06-15
fix: stop mutating package-level columns variable in member list view

### LINKED ISSUE #1038: [bug]: ListMembers mutates global columns variable breaking subsequent --wide calls
## Description
ListMembers in pkg/views/member/list/view.go mutates a package-level `var columns` slice when rendering in non-wide mode. Since it's a global variable, a non-wide call permanently removes columns from it. Every subsequent call — even with --wide — renders an incomplete table.

## Steps to Reproduce
1. Run `harbor project member list --wide` — shows all 6 columns
2. Run `harbor project member list` (non-wide) — shows 4 columns (correct)
3. Run `harbor project member list --wide` again — now shows only 4 columns (broken)

Root cause: line 65 does `columns = utils.RemoveColumns(columns, colsToRemove)` directly on the global.

## Expected Behavior
--wide should always show all 6 columns regardless of previous calls.

## Actual Behavior
After any non-wide call, --wide permanently shows only 4 columns.

## Additional Context
Fix is to make a local copy of the columns slice at function entry and operate on that instead. PR #1002 has the fix ready.

AdeshDeshmukh:
## Description

`ListMembers` in `pkg/views/member/list/view.go` modifies a package-level `var columns` slice when rendering in non-wide mode. Since the variable is global, a non-wide call permanently strips columns from it — meaning subsequent calls with `--wide` will render an incomplete table.

This is a classic shared-mutable-state bug. The fix is straightforward: make a local copy at function entry and operate on that instead. The global now acts as an immutable template.

## Changes

-   `pkg/views/member/list/view.go` — copy `columns` into a local variable, use it for both mutation and the table model

## Testing

-   `go build ./pkg/views/member/...` — passes
-   `go vet ./pkg/views/member/...` — passes
-   Manually verified: `harbor project member list` and `--wide` both render correctly after the change

Fixes #1038

### DIFF
--- a/pkg/views/member/list/view.go
+++ b/pkg/views/member/list/view.go
@@ -37,6 +37,13 @@ var columns = []table.Column{
 
 func ListMembers(members []*models.ProjectMemberEntity, wide bool) {
 	var rows []table.Row
+	localColumns := columns
+
+	if !wide {
+		colsToRemove := []string{"Role ID", "Project ID"}
+		localColumns = utils.RemoveColumns(localColumns, colsToRemove)
+	}
+
 	for _, member := range members {
 		memberID := strconv.FormatInt(member.ID, 10)
 		roleName := utils.CamelCaseToHR(member.RoleName)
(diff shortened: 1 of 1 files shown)


qcserestipy:
This references the wrong issue, please open a dedicated real issue

AdeshDeshmukh:
Hey @qcserestipy — I've opened issue #1038 and updated the PR description to link to it. Also resolved the Copilot threads. Let me know if there's anything else you'd like me to adjust. Thanks!

AdeshDeshmukh:
@bupd PTAL

PR #1007, opened 2026-06-17
Feat: implement OIDC flow

### LINKED ISSUE #1006: [feature]: Harbor CLI OIDC auth (LFX Term-2, June-Augut 2026)
## Problem
`harbor-cli` currently supports username/password login, but it does not support native OIDC login.

This becomes a problem when Harbor is configured to use OIDC as the primary authentication mode. In that setup, users can log in to Harbor through the browser, but `harbor-cli` cannot initiate and complete a comparable OIDC flow.

From a user point of view, the missing capability is:

```sh
harbor login <server> --oidc
```

where the CLI can guide the user through an OIDC login flow and then store a usable credential for subsequent Harbor API calls.

This work is also being pursued as part of **LFX Mentorship Term 2 2026**, so I am actively implementing and iterating on this feature.

---
This issue adds work-in-progress OIDC login support for harbor-cli together with the Harbor Core changes needed to support that flow.

Current implementation in #1007 & [Habor - #24421](https://github.com/goharbor/harbor/pull/23421) allows harbor-cli to start an OIDC login through Harbor Core, return a browser login URL, poll Harbor Core for completion using OIDC state, store the returned credential locally, and use bearer authentication for later API requests.

This change is needed because Harbor currently supports browser-based OIDC login, but harbor-cli does not yet have a native OIDC login flow. The current implementation is still being actively worked on and is not final yet.

NishchayRajput:
## Description
This pull request adds work-in-progress OIDC login support for `harbor-cli` together with the Harbor Core changes needed to support that flow.

The current implementation allows `harbor-cli` to start an OIDC login through Harbor Core, return a browser login URL plus a CLI-only polling token, poll Harbor Core for completion, store the returned credential locally, and use bearer authentication for later API requests.

This change is needed because Harbor currently supports browser-based OIDC login, but `harbor-cli` does not yet have a native OIDC login flow. The current implementation is still being actively worked on and is not final yet.

- Fixes #1006 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
- Added `harbor login <server> --oidc`
- Added Harbor Core support for `GET /c/oidc/login?mode=cli`
- Added Harbor Core polling endpoint for CLI login completion:
  `GET /c/oidc/cli-token?poll_token=...`
- Added cache-backed CLI OIDC login state handling in Harbor Core
- Added local OIDC credential storage support in `harbor-cli`:
  - `auth-type`
  - `id-token`
  - `refresh-token`
  - `expires-at`
- Added bearer-token client path for OIDC credentials
- Added CLI and Harbor test coverage for the OIDC polling flow
- Updated the design/plan documentation to reflect the current polling-token-based flow

## Flow Diagram
image

## Video
[Screencast from 21-06-26 04:07:57 PM IST.webm](https://github.com/user-attachments/assets/886ce52c-0d34-41f0-974a-686598f83e98)
### In case of token expiry cli gives output as 
image


NishchayRajput:
@bupd Please have look at the idea. We may also required changes on harbore-core using existing oidc flow still required still needed few changes like polling state to get the id_token and other details. It's not completed yet but wanted you to have a look and design level decisions. 

PS: ignore lint for now :)

NishchayRajput:
Suggestion:
- We can use URL shortener  or copy support for the  link in the terminal itself.
- Also should display the QR code for the url incase user can scan via phone.

Please let me know if you have more suggestion.

qcserestipy:
@NishchayRajput Had a brief look through this. It looks really really cool!
Just one question out of interest. After the oidc login the id_token is retrieved for identity info.
The refresh_token is also stored for refreshing which is really nice. Is there already functionality included to make use of the refresh_token anyhow or is this just added for completenes currently?
Should we also store the access_token?

Lastly, which party is validating the trust relationship to the oidc source. Just to confirm: this is governed through OIDC providers that are configured in harbor itself and we just make use of the trusted identity provider from Harbor itself.

NishchayRajput:
> @NishchayRajput Had a brief look through this. It looks really really cool! Just one question out of interest. After the oidc login the id_token is retrieved for identity info. The refresh_token is also stored for refreshing which is really nice. Is there already functionality included to make use of the refresh_token anyhow or is this just added for completenes currently? Should we also store the access_token?
> 
> Lastly, which party is validating the trust relationship to the oidc source. Just to confirm: this is governed through OIDC providers that are configured in harbor itself and we just make use of the trusted identity provider from Harbor itself.

Hey, Thanks @qcserestipy for review.
Below is how flow would look like 
image

NishchayRajput:
https://github.com/user-attachments/assets/f8e6a12b-36e3-4c01-a577-cb3e878ccddc

Added Refresh Token support.

pkg/utils/client.go
```diff
@@ -95,3 +123,207 @@ func GetClientByCredentialName(credentialName string) (*v2client.HarborAPI, erro
...
+
+func oidcTokenNeedsRefresh(credential Credential, idToken string) bool {
+	expiresAt, err := oidcTokenExpiryUnix(idToken)
+	if err != nil {
+		expiresAt = credential.ExpiresAt
+	}
+	if expiresAt <= 0 {
+		return false
+	}
+	return time.Now().Unix() >= expiresAt-60
```
qcserestipy, line 183: Why -60 here?

Vad1mo:
not a big fan of 

> added harbor login <server> --oidc

is that the only option to have attach --oidc?
Can this not be derived automatically?
In what situations would you not want to have it automatically and manually specify it?

NishchayRajput:
> not a big fan of
> 
> > added harbor login  --oidc
> 
> is that the only option to have attach --oidc? Can this not be derived automatically? In what situations would you not want to have it automatically and manually specify it?

Yes this can be derived automatically via `/api/v2.0/systeminfo` endpoint, this would allow harbor login to automatically initiate the OIDC flow when the target Harbor instance is configured for OIDC, without requiring the user to explicitly specify `--oidc`.

But besides the oidc flow we also support auth via `cli_secret`, so in those cases simply knowing `auth_mode` doesn't necessarily indicate the authentication flow the user intends to use.

Best way can be interactive solution which was discussed in this weekly meet that we can have over `harbor login`, user can choose the auth method.

cc: @Vad1mo

Vad1mo:
> But besides the oidc flow we also support auth via `cli_secret`, so in those cases simply knowing `auth_mode` doesn't necessarily indicate the authentication flow the user intends to use.
> 
> Best way can be interactive solution which was discussed in this weekly meet that we can have over `harbor login`, user can choose the auth method.

can we have this flow. 

keep --oidc etc. 

1. if no value provided try to determin automatically 
2. if arg provided use that explicitly

One important not on **interactive solution**. Make sure **interactive** is only one of the option, it should work across CI/CD and AI agents using the CLI...

NishchayRajput:
> can we have this flow.
> 
> keep --oidc etc.
> 
> 1. if no value provided try to determin automatically
> 2. if arg provided use that explicitly
> 
> One important not on **interactive solution**. Make sure **interactive** is only one of the option, it should work across CI/CD and AI agents using the CLI...


Yes indeed. Just to summarize the proposed flow:

1. If authentication flags are provided (e.g. --oidc, -u <username> -p <password>, etc.), use the explicitly requested authentication flow.
2. If no authentication flags are provided:
- Query the /api/v2.0/systeminfo endpoint to determine the configured auth_mode.
- If auth_mode is oidc_auth, the CLI can offer an interactive choice between the supported OIDC authentication methods (e.g. browser-based OAuth or CLI secret-based authentication, if supported).
- If auth_mode is db_auth, continue with the existing interactive username/password login flow.


For AI Agents and CI/CD path flags paths remains covered.
cc : @Vad1mo @bupd

pkg/utils/oidc.go
```diff
@@ -0,0 +1,247 @@
...
+	endpoint, err := joinServerPath(serverAddress, oidcCLITokenPath)
+	if err != nil {
+		return nil, err
+	}
+	u, err := url.Parse(endpoint)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse OIDC token endpoint: %w", err)
+	}
+	q := u.Query()
+	q.Set("poll_token", pollToken)
```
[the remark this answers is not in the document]
NishchayRajput: Using poll_token instead of state as state is exposed in the url and anyone can get the session via that state.

pkg/utils/client.go
```diff
@@ -95,3 +123,207 @@ func GetClientByCredentialName(credentialName string) (*v2client.HarborAPI, erro
...
+		return "", fmt.Errorf("failed to get current Harbor data: %w", err)
+	}
+	if err := UpdateOIDCTokens(credential.Name, refreshResp.IDToken, nextRefreshToken, refreshResp.ExpiresAt, harborData.ConfigPath); err != nil {
+		return "", fmt.Errorf("failed to persist refreshed OIDC tokens: %w", err)
+	}
+
+	return refreshResp.IDToken, nil
+}
+
+func oidcTokenNeedsRefresh(credential Credential, idToken string) bool {
```
qcserestipy, line 174: when the jwt cant be parsed this silently falls back to credential.ExpiresAt. maybe add a debug log so refresh issues are easier to spot.

pkg/utils/oidc.go
```diff
@@ -0,0 +1,251 @@
...
+			return nil, fmt.Errorf("OIDC refresh failed: %s", refreshResp.Error)
+		}
+		fallthrough
+	default:
+		body, _ := io.ReadAll(io.LimitReader(resp.Body, 1024))
+		return nil, fmt.Errorf("failed to refresh OIDC token: status %d: %s", resp.StatusCode, string(body))
+	}
+}
+
+func joinServerPath(serverAddress, path string) (string, error) {
```
qcserestipy, line 278: i think the copilot comment above is a false positive, this keeps u.Path and appends, so a reverse-proxied /harbor stays intact. maybe add a small test with a base-path server url to close it out.

qcserestipy (changes requested):
Thank you for your contribution! nice work, the polling flow looks solid and the recent fixes (mutex guarding, the 202/timeout test) look good. just two small things left below.

pkg/utils/client.go
NishchayRajput: refreshing around expiration ( when 1 min is left ).

pkg/utils/client.go
NishchayRajput: done!

bupd (changes requested):
`--auth-mode` this Should be the flag instead of OIDC. 
flag options
1. oidc
2. ldap
3. db (usually used in the case of `username: admin`- sole system admin) 

without giving this flag should fallback first to oidc > db. 

image

image

image

NishchayRajput:
```
Error: failed to decode OIDC login response: invalid character '<' looking for beginning of value
exit status 1
```
- Above error is expected as currently you don't use the corresponding changes made in [HARBOR-PR#23421](https://github.com/goharbor/harbor/pull/23421). Since while triggering Harbor OIDC mode from CLI, without those changes would trigger oidc auth flow and return redirect html output instead of JSON format output as expected by CLI. (Will add a check for json response and will give proper error output on CLI side)

- Also for the second thing
```go run ./cmd/harbor/main.go login registry.goharbor.io -u bupdprasanth@gmail.com```
This is shouldn't ideally go to oidc mode as user may user cli_secret so we have to support that as well. Also it may be possible that admin may login via idp.

@bupd

NishchayRajput:
This image shows output for --auth-mode oidc
command: `harbor login --auth-mode oidc http://localhost:4200 -v`
image


When user tries to login when auth mode is db
command: `harbor login --auth-mode oidc http://localhost:4200 -v`
image

bupd (changes requested):
@NishchayRajput  not working yet,

please check and verify against https://registry.goharbor.io 

Image

NishchayRajput:
> @NishchayRajput  not working yet,
> 
> 
> 
> please check and verify against https://registry.goharbor.io 
> 
> 
> 
> Image

To login via oidc there's need of harbor changes, which are present in pr #23421 on harbor

bupd:
why shouldnt this work out of the box in harbor-cli ??

bupd:
I dont think we need change from the harbor core side.

NishchayRajput:
Using the existing OIDC flow as-is is not straightforward here. When Harbor CLI opens the browser, it has no visibility into the browser's auth state. The browser communicates directly with the IdP, and the IdP then communicates with harbor-core. Harbor CLI is completely out of that loop and has no way to know when authentication has completed.

To support this flow, Harbor CLI would need to poll Harbor to retrieve the auth state. That is the minimum change required if we want to reuse the existing OIDC flow in Harbor.

and This is exactly what I have done.

NishchayRajput:
> Using the existing OIDC flow as-is is not straightforward here. When Harbor CLI opens the browser, it has no visibility into the browser's auth state. The browser communicates directly with the IdP, and the IdP then communicates with harbor-core. Harbor CLI is completely out of that loop and has no way to know when authentication has completed.
> 
> To support this flow, Harbor CLI would need to poll Harbor to retrieve the auth state. That is the minimum change required if we want to reuse the existing OIDC flow in Harbor.
> 
> and This is exactly what I have done.

If you have any other idea/opinion please let me know.

bupd:
okay cool, can you check and add some tests to ensure that is backward compatible and does it needs update in the ui of harbor.

NishchayRajput:
> okay cool, can you check and add some tests to ensure that is backward compatible and does it needs update in the ui of harbor.

Sure. Also we don't require any UI changes.

PR #1008, opened 2026-06-19
feat(schedule): add smart router for schedule create and delete commands

### LINKED ISSUE #1009: [feature]:Add create and delete commands for system schedules
## Problem
What problem does this feature solve?  
Describe the limitation or pain point you are facing.

Currently, the `harbor-cli` only supports listing existing schedules (`harbor schedule list`). There is no way to automate the creation or deletion of critical scheduled jobs like Garbage Collection (GC) or Vulnerability Scanning (Scan-All) directly from the terminal. Users are forced to fall back to the Harbor Web UI or manually construct raw HTTP REST calls to manage their system schedules.

## Proposed Solution
Describe the feature or behavior you would like to see.

I propose adding `create` and `delete` subcommands under the `harbor schedule` namespace. 

Since Harbor's Go client SDK requires hitting different endpoints for different schedule types (e.g., `client.GC.UpdateGCSchedule` vs `client.ScanAll.UpdateScanAllSchedule`), the CLI should implement a unified routing layer. Users should be able to pass a `--type` flag alongside a `--cron` string. The router will construct the generic JSON payload and dynamically route it to the correct Harbor SDK client.

**Expected Usage:**
```bash
# Create a schedule
harbor schedule create --type gc --cron "0 0 0 * * *"

# Delete a schedule
harbor schedule delete --type gc

Psykii22:
## Description
This pull request introduces the ability to create and delete scheduled jobs directly from the CLI. 

Previously, the `harbor schedule` command only supported listing existing schedules. Because Harbor's Go client SDK has completely separate services for updating each job type (e.g., `client.GC`, `client.ScanAll`), this PR implements a unified "Smart Router" pattern in `pkg/api/schedule_handler.go`. 

This pattern allows users to pass a generic `--type` flag from a single CLI command and routes the schedule payload to the correct Harbor backend API automatically. **For this initial PR, we have implemented routing for the two most common system-wide schedules (`gc` and `scan-all`). The router is designed to be easily extensible so that the remaining schedule types (like `purgeaudit`, `replication`, etc.) can be seamlessly added in future PRs.**

Fixes #1009 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
- Created `pkg/api/schedule_handler.go` to act as a unified router for constructing schedule payloads and sending them to the correct Harbor endpoints.
- Added `harbor schedule create` command, requiring a `--type` (e.g., `gc`, `scan-all`) and a 6-field `--cron` string.
- Added `harbor schedule delete` command, which sends a schedule payload of type `None` to the specified job type to disable it.
- Registered the new subcommands in `cmd/harbor/root/schedule/cmd.go`.

## Examples & Usage
Here are the new commands that manage schedules, and examples of what the schedule format looks like:

**1. Create a Garbage Collection schedule:**
`harbor schedule create --type gc --cron "0 0 * * * *"`

**2. Create a Scan-All schedule:**
`harbor schedule create --type scan-all --cron "0 0 * * * *"`

**3. Delete schedules:**
`harbor schedule delete --type gc`
`harbor schedule delete --type scan-all`

### Schedule Format:
The `--cron` flag expects a 6-field cron string (which Harbor requires instead of the standard 5-field linux cron). The format is:
`Seconds Minutes Hours Day-of-month Month Day-of-week`

When a user runs the `delete` command, the CLI automatically passes `Type: "None"` to the backend Harbor API to disable the schedule.

### DIFF
--- a/cmd/harbor/root/schedule/create.go
+++ b/cmd/harbor/root/schedule/create.go
@@ -1,0 +1,53 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schedule
+
+import (
+	"fmt"
+
+	"github.com/goharbor/harbor-cli/pkg/api"
+	"github.com/spf13/cobra"
+)
+
+func CreateScheduleCommand() *cobra.Command {
+	var jobType string
+	var cronString string
+
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Create a schedule job (e.g. gc, scan-all)",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if jobType == "" {
+				return fmt.Errorf("--type flag is required")
+			}
+			if cronString == "" {
+				return fmt.Errorf("--cron flag is required")
+			}
+
+			err := api.CreateSchedule(jobType, cronString)
+			if err != nil {
+				return fmt.Errorf("failed to create %s schedule: %v", jobType, err)
+			}
+
+			fmt.Printf("Successfully created schedule for %s with cron: %s\n", jobType, cronString)
+			return nil
+		},
+	}
+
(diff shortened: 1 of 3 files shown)


NucleoFusion:
I like the idea of this, please make an issue and connect the PR to that issue for this.
And explain in detail why this is nice and what is required for us to do this cleanly.

NucleoFusion:
Also, if you could, mention all the commands that have schedule, and what the schedule looks like

Psykii22:
Should i add commands for other schedules types also like the purgeaudit and replication. and should i open new issues for that or add in this one

NucleoFusion:
Not yet, it should be a separate PR, but we should discuss the pattern, and if we want any modifications etc

PR #1014, opened 2026-06-22
fix(project): handle registry list retrieval error gracefully

### LINKED ISSUE #1040: fix(project): handle registry list retrieval error gracefully in create view
## Problem
In `pkg/views/project/create/view.go`, if `getRegistryList()` fails, the entire project creation form is blocked — even though the registry selector is only needed when "Proxy Cache" is enabled. There are also unnecessary nil checks before ranging over `registries.Payload`.

## Proposed Solution
- Treat the registry list as optional: log a warning and continue if retrieval fails
- Remove unnecessary nil checks (ranging over a nil slice in Go is safe)
- Add a comment making the non-fatal intent explicit

## Affected File
- `pkg/views/project/create/view.go`

saiashok0981:
Closes #1040
### Summary
This PR makes interactive project creation resilient to registry listing API failures.

### Rationale & Design Decisions
- Currently, if the `Registry` listing endpoint fails or is unauthorized, `CreateProjectView` fails immediately.
- Listing registries is only required for selecting a proxy cache registry.
- By handling the error gracefully (logging a debug message and passing an empty list of options), standard project creation remains functional for users without administrative registry access.
Signed-off-by: saiashok103@gmail.com

### DIFF
--- a/pkg/views/project/create/view.go
+++ b/pkg/views/project/create/view.go
@@ -53,15 +53,17 @@ func getRegistryList() (*registry.ListRegistriesOK, error) {
 
 func CreateProjectView(createView *CreateView) error {
 	theme := huh.ThemeCharm()
-	// I want it to be a map of registry ID to registry name
+	// Registry list is optional: if retrieval fails, the proxy-cache registry
+	// selector will simply be hidden and the form can still proceed.
+	registryOptions := map[string]string{}
 	registries, err := getRegistryList()
 	if err != nil {
-		return err
-	}
-	registryOptions := map[string]string{}
-	for _, registry := range registries.Payload {
-		regiId := fmt.Sprintf("%d", registry.ID)
-		registryOptions[regiId] = fmt.Sprintf("%s (%s)", registry.Name, registry.URL)
+		log.Warnf("failed to retrieve registry list, proxy cache registry selection will be unavailable: %v", err)
+	} else {
+		for _, r := range registries.Payload {
+			regiId := fmt.Sprintf("%d", r.ID)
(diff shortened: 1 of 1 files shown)


qcserestipy:
Looking at both getRegistryList() and ListRegistries(), the nil checks here seem unnecessary. Both wrappers return nil, err on failure and response, nil on success. If the generated client could theoretically return nil, nil, that should probably be guarded once in the wrapper rather than in every caller.

Also, checking Payload != nil is not needed because ranging over a nil or empty slice is safe in Go.

The actual behavior change is that errors from getRegistryList() are no longer returned. If the registry list is optional for rendering this view, logging and continuing makes sense, but we should make that intention explicit.

qcserestipy:
Also DCO is failing and this PR has no linked issue

saiashok0981:
Hi @qcserestipy, thank you for the detailed review! I've addressed all the points:

Removed unnecessary nil checks:  Dropped the registries != nil and registries.Payload != nil guards. Since getRegistryList() already returns nil, err on failure and response, nil on success, these checks were redundant. Ranging over a nil slice in Go is also safe.

Made the intention explicit: The registry list is optional for this view. If retrieval fails, the proxy cache registry selector is simply hidden and the form proceeds. I've added a comment to make this clear and changed the error handling to log.Warnf(...) instead of returning the error.

DCO fixed: git user.name was not configured globally, causing a malformed Signed-off-by. Amended the commit with the correct sign-off and force-pushed.

I still need to link an issue will do that shortly. Please re-review when you get a chance!

PR #1016, opened 2026-06-22
feat: add logout command

### LINKED ISSUE #1019: [feature]: logout command
## Problem
Harbor CLI has a `login` command but no `logout` command 

## Proposed Solution
Add a harbor logout command that:
- Removes the current credential from the local config file
- Asks for confirmation before logging out (--yes flag to skip)
- Shows "Not logged in." if no active session exists

Sypher845:
## Description
Adds a harbor logout command for better UX, gives users a simple, discoverable way to remove their current credentials from the local config, mirroring the standard login / logout pairing found in CLI tools. Includes a confirmation prompt with -y flag.

- Fixes #1019 

## Command
- `harbor logout` 
Kooha-2026-06-23-02-04-56

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance


NucleoFusion (commented):
Is this not what 
`harbor-cli context delete` does?
And `context delete` also has support for providing a name?

Sypher845:
Yeah, this is mainly for better UX and discoverability,  most CLI tools have a login/logout pairing

cmd/harbor/root/logout.go
```diff
@@ -0,0 +1,87 @@
...
+			if config.CurrentCredentialName == "" {
+				fmt.Println("Not logged in.")
+				return nil
+			}
+
+			currentName := config.CurrentCredentialName
+
+			if !skipConfirm {
+				fmt.Printf("Log out from '%s'? [y/N]: ", currentName)
+
```
NucleoFusion, line 51: Should be in `prompt` and use `charm/huh`

NucleoFusion (changes requested):
Im still conflicted on whether this is a meaningful enough change yk
But that is for the meet, while that is being done, here is a review anyway

qcserestipy:
@NucleoFusion @Sypher845 I also think that logout can be an alias for context delete at maximum. But it would reimplement the same functionality twice. Making it a convenience proxy might be discussed.

bupd (approved):
/lgtm

PR #1022, opened 2026-06-25
fix(preheat): handle nil payload during policy update

### LINKED ISSUE #1021: [bug]: preheat policy update command panics on empty or nil server payload
## Description
When running the preheat policy update command (`harbor project preheat policy update`), if the Harbor API server returns a successful response but the payload body is nil or empty, the CLI immediately crashes with a nil pointer dereference panic instead of printing a proper error message.

## Steps to Reproduce
1. Run command: `harbor project preheat policy update [projectName] [policyName]`
2. With arguments/config: Execute against a client/server mock where `GetPreheatPolicy` returns an empty payload.
3. Observe error: The CLI command crashes with a panic trace.

## Expected Behavior
The CLI should validate that the retrieved policy and payload are not nil, and if they are, return a clean error message (e.g. `preheat policy <name> payload is empty`) rather than crashing.

## Actual Behavior
The CLI panics with:
`panic: runtime error: invalid memory address or nil pointer dereference`

## Environment
- OS: Windows 11 (also affects macOS/Linux)
- Tool version: Latest main branch
- Other relevant details: Go version go1.21.0+

## Additional Context
In `cmd/harbor/root/project/preheat/policy/update.go`, the command retrieves the policy details:
```go
existingPolicy, err := api.GetPreheatPolicy(projectName, policyName)

### PR DESCRIPTION
**Pull request by shellyco-code**
## Description
This pull request fixes a runtime nil pointer dereference panic in the `harbor project preheat policy update` command when the fetched preheat policy or its payload from the API is nil/empty.

- Fixes #1021

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
- Modified `cmd/harbor/root/project/preheat/policy/update.go` to validate if `existingPolicy` and `existingPolicy.Payload` are nil, returning a clean error instead of panicking.
- Refactored `api.GetPreheatPolicy` call to use a package-level mockable variable `getPreheatPolicyFunc` to enable clean unit testing.
- Created `cmd/harbor/root/project/preheat/policy/update_test.go` to add unit tests mocking nil policy and nil payload API return values.

## Testing
- Formatted modified code using `gofmt -s -w`.
- Compiled and verified the CLI binary builds successfully locally (`go build`).
- Ran preheat policy package tests (`go test ./cmd/harbor/root/project/preheat/policy/...`), which passed successfully.
- Ran the full repository test suite (`go test ./...`), which passed successfully without regressions.

### DIFF
--- a/cmd/harbor/root/project/preheat/policy/update.go
+++ b/cmd/harbor/root/project/preheat/policy/update.go
@@ -26,10 +26,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var (
-	getPreheatPolicyFunc = api.GetPreheatPolicy
-)
-
 func UpdatePolicyCommand() *cobra.Command {
 	var isID bool
 
@@ -79,7 +75,7 @@ func UpdatePolicyCommand() *cobra.Command {
 			}
 
 			log.Debug("Fetching preheat policy...")
-			existingPolicy, err := getPreheatPolicyFunc(projectName, policyName)
+			existingPolicy, err := api.GetPreheatPolicy(projectName, policyName)
 			if err != nil {
 				if utils.ParseHarborErrorCode(err) == "404" {
 					return fmt.Errorf("preheat policy %s not found in project %s", policyName, projectName)
--- a/cmd/harbor/root/project/preheat/policy/update.go
+++ b/cmd/harbor/root/project/preheat/policy/update.go
@@ -26,6 +26,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	getPreheatPolicyFunc = api.GetPreheatPolicy
+)
+
 func UpdatePolicyCommand() *cobra.Command {
 	var isID bool
 
@@ -75,12 +79,16 @@ func UpdatePolicyCommand() *cobra.Command {
 			}
 
 			log.Debug("Fetching preheat policy...")
-			existingPolicy, err := api.GetPreheatPolicy(projectName, policyName)
+			existingPolicy, err := getPreheatPolicyFunc(projectName, policyName)
 			if err != nil {
 				if utils.ParseHarborErrorCode(err) == "404" {
 					return fmt.Errorf("preheat policy %s not found in project %s", policyName, projectName)
 				}
 				return fmt.Errorf("failed to get preheat policy: %v", utils.ParseHarborErrorMsg(err))
+			}
+
+			if existingPolicy == nil || existingPolicy.Payload == nil {
+				return fmt.Errorf("preheat policy %s payload is empty", policyName)
 			}
 
 			log.Debug("Fetching available providers...")
--- a/cmd/harbor/root/project/preheat/policy/update_test.go
+++ b/cmd/harbor/root/project/preheat/policy/update_test.go
@@ -15,10 +15,10 @@
 package policy
 
 import (
-	"bytes"
 	"testing"
 
 	"github.com/goharbor/go-client/pkg/sdk/v2.0/client/preheat"
+	"github.com/goharbor/harbor-cli/pkg/testutil"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: cmd/harbor/root/project/preheat/policy/update.go [hunk 1]
```diff
@@ -26,6 +26,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	getPreheatPolicyFunc = api.GetPreheatPolicy
+)
+
```
NucleoFusion: we generally discourage this method of testing since it makes it kinda complicated to understand.
Our method for testing is to just test the command failures, which is wrong flags, malformed input etc

The full command testing is TBD in the e2e tests which are in-dev by a support feature that @bupd i think is making.

cmd/harbor/root/project/preheat/policy/update_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	getPreheatPolicyFunc = func(projectName, policyName string) (*preheat.GetPolicyOK, error) {
+		return nil, nil
+	}
+
+	cmd := UpdatePolicyCommand()
+	var buf bytes.Buffer
+	cmd.SetOut(&buf)
+	cmd.SetErr(&buf)
+	cmd.SetArgs([]string{"my-project", "my-policy"})
```
NucleoFusion: There is a test helper function, in /pkg/testutil/testutil.go
With function signature,
func TestCmd(t *testing.T, cmdFunc func() *cobra.Command, flags ...string) error {

This takes the cmd & the flags, and just outputs the error (if any)
so you can just use this

NucleoFusion (changes requested):
Testing for cmds is limited to testing for failures. By that I mean that we induce errors by providing malformed input/flags to check whether the command appropriately fails.
By that I mean that providing invalid flags, too large values etc.

The PR resolves the issue, but tests are not up to the general CLI standard, all else looks great!

cmd/harbor/root/project/preheat/policy/update_test.go
shellyco-code: Thanks for the feedback! I've updated the tests to use the existing pkg/testutil.TestCmd helper and aligned them with the Harbor CLI testing conventions. I also verified that the test suite passes after the changes. Please let me know if there's anything else I should adjust.

cmd/harbor/root/project/preheat/policy/update.go
NucleoFusion: We can just remove this now, and the relevant things in tests, cause we dont use it

cmd/harbor/root/project/preheat/policy/update_test.go
```diff
@@ -0,0 +1,53 @@
...
+		getPreheatPolicyFunc = originalGetPreheatPolicy
+	})
+
+	getPreheatPolicyFunc = func(projectName, policyName string) (*preheat.GetPolicyOK, error) {
+		return &preheat.GetPolicyOK{Payload: nil}, nil
+	}
+
+	err := testutil.TestCmd(t, UpdatePolicyCommand, "my-project", "my-policy")
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "payload is empty")
```
NucleoFusion: These tests will fail now
cause we are not using the pr.

Have a look at cmd/harbor/root/project/list_test.go
We test only the cases for flags and stuff like that, we dont test for cases that propagate back errors from a more deeper level package

NucleoFusion (changes requested):

NucleoFusion:
DCO is failing btw
have a look at https://github.com/goharbor/harbor-cli/pull/1022/checks?check_run_id=83384485147

cmd/harbor/root/project/preheat/policy/update_test.go
shellyco-code: Thanks for the detailed feedback! That makes sense. I'll remove the `getPreheatPolicyFunc` indirection and update the tests to follow the Harbor CLI testing conventions, using `list_test.go` as a reference. I'll push an updated revision shortly. Thanks!

NucleoFusion (approved):
lgtm! thanks for the contribution!

NucleoFusion (approved):
lgtm!
Thanks for the contribution!

bupd (approved):
/lgtm

PR #1032, opened 2026-06-30
fix(info): prevent nil pointer panic when volumes payload is absent

### LINKED ISSUE #1031: [bug]: harbor info panics with nil pointer dereference when storage volume payload is absent
## Describe the bug

Running `harbor info` can panic with a nil pointer dereference when the Harbor server returns a `SystemInfo` response without storage information.

The code assumes the `SystemInfo` pointer is always non-nil before accessing the `Storage` field. If the API returns a nil `SystemInfo` pointer, the command crashes instead of handling the missing data gracefully.

## Steps to Reproduce

1. Connect Harbor CLI to a Harbor server that returns a `SystemInfo` response with a nil or missing storage information.
2. Run:

```bash
harbor info
```

3. Observe the panic.

## Actual Behavior

The command terminates with:

```text
panic: runtime error: invalid memory address or nil pointer dereference
```

## Expected Behavior

`harbor info` should not panic when storage information is unavailable. Instead, it should continue executing and display default or empty values for the storage-related fields.

## Location

File:
```
pkg/views/info/list/view.go
```

Function:
```
CreateSystemInfo(...)
```

Relevant lines:
- Around line 74
- Around line 80

## Root Cause

The implementation checks whether `volumes.Storage` is nil before accessing the storage slice, but it never verifies that the `volumes` pointer itself is non-nil.

When the API returns a nil `SystemInfo` pointer, dereferencing `volumes.Storage` causes a nil pointer dereference.

## Suggested Fix

Guard against a nil `SystemInfo` pointer before accessing the `Storage` field so the command handles missing storage information gracefully instead of panicking.

shellyco-code:
## Description

This PR fixes a nil pointer dereference in `harbor info` when the Harbor server returns a `SystemInfo` response without storage information.

The current implementation assumes the `volumes` pointer is always non-nil before accessing `volumes.Storage`, which can lead to a runtime panic when the API returns an incomplete response.

This change adds a nil guard before accessing the storage information so the command returns zero values instead of crashing.

Fixes #1031

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes

- Added a nil check before accessing `volumes.Storage`.
- Prevented a nil pointer dereference when `SystemInfo.Volumes` is nil.
- Preserved existing behavior for valid responses.

## Testing

- [x] Verified the code path.
- [x] Existing behavior remains unchanged for valid responses.

### DIFF
--- a/pkg/views/info/list/view.go
+++ b/pkg/views/info/list/view.go
@@ -71,13 +71,13 @@ func CreateSystemInfo(
 		SystemInfo: generalInfo,
 		VolumeInfo: &Volumes{
 			Free: func() uint64 {
-				if volumes.Storage != nil && len(volumes.Storage) > 0 {
+				if volumes != nil && volumes.Storage != nil && len(volumes.Storage) > 0 {
 					return volumes.Storage[0].Free
 				}
 				return 0
 			}(),
 			Total: func() uint64 {
-				if volumes.Storage != nil && len(volumes.Storage) > 0 {
+				if volumes != nil && volumes.Storage != nil && len(volumes.Storage) > 0 {
 					return volumes.Storage[0].Total
 				}
 				return 0
--- a/pkg/views/info/list/view_test.go
+++ b/pkg/views/info/list/view_test.go
@@ -1,0 +1,66 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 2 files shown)


NucleoFusion (approved):
lgtm!
Thanks for the contribution!

PR #1050, opened 2026-07-14
[feat]: added v2 models for loading state

NishchayRajput:
## Description

Adds async loading-state support for the `harbor artifact list` interactive flow so the terminal no longer appears frozen while data is being fetched. This introduces v2 base Bubble Tea models for async `table`, `selection`, `multiselect`, and `grid lifecycles`, and uses them to move `artifact list` fetching and `project/repository/artifact` prompt loading into the TUI lifecycle.
This change is needed because the previous interactive path fetched data before the TUI rendered, which left users staring at a blank or static terminal with no feedback during slow API calls.

- Fixes #

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
- Added async v2 base models for tablelist, selection, multiselect, and tablegrid with loading, ready, and error states.
- Updated harbor artifact list to keep formatted output (json|yaml|csv) on the existing direct-fetch path and use the new async TUI path only for interactive output.
- Moved project, repository, and artifact selection loading into the interactive lifecycle so loading feedback appears on the existing terminal screen instead of looking frozen.

## Demo
[Screencast from 14-07-26 08:27:02 AM IST.webm](https://github.com/user-attachments/assets/6d6d3106-4f78-47c3-89ec-c04c01694017)


NucleoFusion (commented):
Could you add usage for selection/multiselect/tablegrid so we could see how it looks/works?
maybe find a new name for `presenter` as well, i think we discussed changing this since this doesnt really give much info about the package

NishchayRajput:
> Could you add usage for selection/multiselect/tablegrid so we could see how it looks/works? maybe find a new name for `presenter` as well, i think we discussed changing this since this doesnt really give much info about the package

artifact list already use the `tablelistv2` and `selectionv2` models, for `multiselectv2` and `tablegridv2` models please find attached.


https://github.com/user-attachments/assets/39c7a823-46ba-4e6b-8f39-81dc8d733fa1


https://github.com/user-attachments/assets/bdf8825c-a9cc-44cd-847c-aba07c967e5a

@bupd @NucleoFusion @qcserestipy

NucleoFusion:
I want to see the code, on how the new package would look like for these models
If theres even some change.
You can give me the link to one of your branches which may have it

NishchayRajput:
> I want to see the code, on how the new package would look like for these models If theres even some change. You can give me the link to one of your branches which may have it

Already pushed in this PR. Idea remains same passing loadingdata function in the model itself.

NucleoFusion:
> > I want to see the code, on how the new package would look like for these models If theres even some change. You can give me the link to one of your branches which may have it
> 
> Already pushed in this PR. Idea remains same passing loadingdata function in the model itself.

Oh okay thanks for that, will have a look

pkg/views/robot/select/view.go
```diff
@@ -103,64 +106,157 @@ func NewRobotPermissionsGrid(kind string) (*tablegrid.TableGrid, error) {
...
+				choices = append(choices, *permission)
+			}
+
+			if len(choices) == 0 {
+				return nil, fmt.Errorf("no %s permissions found", kind)
+			}
+
+			return choices, nil
+		},
+	)
```
NucleoFusion: This is not how its supposed to look
Like the tablelistv2, we use the new package to define the function and that function is used here....
prompt should not concern with making api calls and data decisions.

NucleoFusion (changes requested):
Pls move the relevant functions to the `presenter` folder (or the appropriate name)
And then only we will be able to ascertain how the repo would look _after_ the merge and changes required

NucleoFusion (commented):
Can you also add the respectice changes to the command? so that I can test it out and see how the other models look from the commands' perspective

NishchayRajput:
> Can you also add the respectice changes to the command? so that I can test it out and see how the other models look from the commands' perspective

We don't have the permission cmd for robot. So basically how earlier flow is robot cmd like create/update/list etc uses prompt.go which than uses view pkgs corresponding to robot. So just to eliminated it, I just used presenter pkg inside the prompt/. May be we can eliminate calling prompt/ inside the `cmd/harbor/root/robot/<function>`. We can directly call presenter pkg inside cmd/harbor/root/robot/.

pkg/views/base/tablelistv2/model.go
```diff
@@ -0,0 +1,166 @@
...
+	"golang.org/x/term"
+)
+
+type state string
+
+const (
+	stateLoading state = "loading"
+	stateReady   state = "ready"
+	stateError   state = "error"
+)
```
NucleoFusion, line 34: this is being reused everywhere
we can move this somewhere..
Including the dataLoaded & dataFailed etc

NucleoFusion (commented):
And the robot command is still using the prompt package
and not the presenter package...

NucleoFusion:
I want it so that every command only talks to the presenter package _or_ the api (for json/yaml output) and NEVER the prompt package.
Basically we are adding another step before the current prompt package so the command now flows like,

cmd -> presenter -> prompt -> model -> on-screen

NishchayRajput:
> I want it so that every command only talks to the presenter package _or_ the api (for json/yaml output) and NEVER the prompt package. Basically we are adding another step before the current prompt package so the command now flows like,
> 
> cmd -> presenter -> prompt -> model -> on-screen

Yes, Ideally now it should be. Earlier prompt package serves as common util store for many commads.

pkg/views/base/tablelistv2/model.go
NishchayRajput: DataLoaded can't be done as it's type interface and is different for all v2 models

PR #1071, opened 2026-07-27
fix(cli): validate arg count for artifact tags create/delete to preve…

### LINKED ISSUE #1069: [bug]: artifact tags create and delete panic with index out of range when exactly one argument is provided
## Description

`harbor artifact tags create` and `harbor artifact tags delete` both panic with an index out of range error when called with exactly one argument.

Both commands support two modes:
- **0 arguments** — launches interactive prompts to select project, repo, reference, and tag
- **2 arguments** — takes `<reference>` and `<tag>` directly, skips prompts

Neither command defines an `Args` validator on the `cobra.Command`. When a user passes exactly one argument, the code enters the non-interactive branch (because `len(args) > 0` is true) and immediately accesses `args[1]`, which does not exist. The CLI crashes instead of showing a usage error.

## Steps to Reproduce

1. Run `harbor artifact tags create myproject/myrepo/sha256:abc123` (provides a reference but omits the tag name).
2. Alternatively, run `harbor artifact tags delete myproject/myrepo/sha256:abc123`.
3. The CLI panics with a runtime index out of range error instead of exiting cleanly.

## Expected Behavior

A clear error message explaining what is missing, printed by Cobra before any code in `RunE` executes:

```text
Error: requires both <project>/<repository>/<reference> and <tag>, got only one argument
```

The CLI should exit with a non-zero status code and make no API calls.

## Actual Behavior

The CLI crashes with:

```text
panic: runtime error: index out of range 1 with length 1
goroutine 1 running:
[github.com/goharbor/harbor-cli/cmd/harbor/root/artifact/tags.(*CreateTagsCmd...).RunE](https://github.com/goharbor/harbor-cli/cmd/harbor/root/artifact/tags.(*CreateTagsCmd...).RunE)(...)
	cmd/harbor/root/artifact/tags/create.go:42
```

The crash originates at these lines in both `cmd/harbor/root/artifact/tags/create.go:36-42` and `delete.go:34-40`:

```go
if len(args) > 0 {
    projectName, repoName, reference, err = utils.ParseProjectRepoReference(args[0])
    if err != nil {
        return fmt.Errorf("failed to parse project/repo/reference: %v", err)
    }
    tagName = args[1]  // ← PANICS when len(args) == 1
} else {
    // interactive prompt path
}
```

## Environment

- **OS:** macOS / Linux / Windows (any)
- **Harbor CLI version:** `main` branch
- **Harbor server version:** N/A — the panic occurs locally before any API call is made

## Additional Context

**Why the simple fix won't work** — The obvious fix, adding `Args: cobra.ExactArgs(2)`, is incorrect here because both commands also support a 0-argument interactive mode. Using `ExactArgs(2)` would break the interactive prompt flow that already works correctly.

**Proposed approach** — Add a custom `Args` validator function that explicitly handles the argument count cases:
- `0 args`: Allowed (launches interactive mode)
- `2 args`: Allowed (non-interactive execution)
- `1` or `3+ args`: Reject with a clear Cobra error message

This preserves the existing interactive functionality while preventing the panic and giving the user a clear, helpful error message.

**Scope** — Both `create.go` and `delete.go` in `cmd/harbor/root/artifact/tags/` have the identical bug. A single pull request will fix both files, as they are in the same package (`artifacttags`) and share the same code pattern.

I believe a custom Args validator could work here , but I'm open to
other approaches if you have a different preference.

mantrapatel05:
## Description
harbor artifact tags create and delete both panic with an "index out of range" error 
when called with exactly one argument. Neither command had an Args validator on the 
cobra.Command, so passing a single arg fell into the non-interactive branch and tried 
to read args[1], which doesn't exist.

- Fixes #1069

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
- Added an Args validator to CreateTagsCmd that allows 0 args (interactive mode) or 2 args (non-interactive), rejects anything else with a clean error
- Added the same Args validator to DeleteTagsCmd
- Verified manually: 1 arg → clean error, no panic; 2 args → works as before; 0 args → interactive prompts still work; go build ./... passes

### DIFF
--- a/cmd/harbor/root/artifact/tags/create.go
+++ b/cmd/harbor/root/artifact/tags/create.go
@@ -28,6 +28,12 @@ func CreateTagsCmd() *cobra.Command {
 		Use:     "create",
 		Short:   "Create a tag of an artifact",
 		Example: `harbor artifact tags create <project>/<repository>/<reference> <tag>`,
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 0 || len(args) == 2 {
+				return nil
+			}
+			return fmt.Errorf("requires either 0 arguments(interactive mode) or 2 arguments(<project>/<repository>/<reference> <tag>), got %d", len(args))
+		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			var err error
 			var projectName, repoName, reference string
--- a/cmd/harbor/root/artifact/tags/delete.go
+++ b/cmd/harbor/root/artifact/tags/delete.go
@@ -27,6 +27,12 @@ func DeleteTagsCmd() *cobra.Command {
 		Use:     "delete",
 		Short:   "Delete a tag of an artifact",
 		Example: `harbor artifact tags delete <project>/<repository>/<reference> <tag>`,
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 0 || len(args) == 2 {
+				return nil
+			}
+			return fmt.Errorf("requires either 0 arguments (interactive mode) or 2 arguments (<project>/<repository>/<reference> <tag>), got %d", len(args))
+		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			var err error
 			var projectName, repoName, reference string


qcserestipy (changes requested):
Thank you for your contribution! lgtm

qcserestipy (approved):

PR #1075, opened 2026-08-01
fix(project): warn when --page is ignored due to --page-size=0

### LINKED ISSUE #1074: [bug]: --page flag silently ignored when --page-size=0 in project list
## Description
In `fetchProjects` (cmd/harbor/root/project/list.go), when --page-size is 0 
(fetch-all mode), the code silently overwrites the user-provided --page value 
to 1 without any warning.

## Steps to Reproduce
harbor project list --page 5 --page-size 0

## Expected
Either warn the user that --page is ignored in fetch-all mode, or return 
an error if both are explicitly set to conflicting values.

## Actual
--page is silently discarded, no feedback given.

## Proposed Fix
Add a log.Warn or validation when opts.Page != 1 (default) and opts.PageSize == 0.

A-d-i-t-y:
Closes #1074

## Problem
When `--page-size=0` (fetch-all mode), the code silently resets `--page` to 1
without informing the user, causing confusing behavior when a user explicitly
passes a `--page` value alongside `--page-size=0`.

## Fix
Added a warning log when `opts.Page != 1` and `opts.PageSize == 0`, informing
the user that `--page` is being ignored.

## Testing
Tested against demo.goharbor.io:
\`harbor project list --page 5 --page-size 0 -v\`

Output now correctly shows:
\`level=warning msg="--page is ignored when --page-size is 0 (fetching all pages)"\`

### DIFF
--- a/cmd/harbor/root/project/list.go
+++ b/cmd/harbor/root/project/list.go
@@ -128,6 +128,9 @@ func ListProjectCommand() *cobra.Command {
 func fetchProjects(listFunc func(...api.ListFlags) (project.ListProjectsOK, error), opts api.ListFlags) ([]*models.Project, error) {
 	var allProjects []*models.Project
 	if opts.PageSize == 0 {
+		if opts.Page != 1 {
+			log.Warn("--page is ignored when --page-size is 0 (fetching all pages)")
+		}
 		log.Debug("Page size is 0, will fetch all pages")
 		opts.PageSize = 100
 		opts.Page = 1


qcserestipy:
Hi @A-d-i-t-y, thank you for taking this on, and apologies for the slow response. After thinking about it, I have decided to close this one. The underlying behavior in #1074 is intentional: --page-size defaults to 0, which means fetch-all mode, and in that mode --page simply has no meaning. Anyone who wants actual pagination can pass an explicit --page-size.

PR #757, opened 2026-03-19
feat: Migrate to Charmbracelet Bubbletea and Bubbles libraries to v2.

marwan562:
## Description
Upgrade CLI to support bubbletea v2

- Closes #756

## Type of Change
Please select the relevant type.

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation update
- [x] Chore / maintenance

PR #1076, opened 2026-08-02
Bubbletea v1 to v2 migration

NishchayRajput:
## Description
The goal is to preserve current Harbor CLI behavior while aligning the codebase with the Charm v2 ecosystem.
The migration affects four main areas:

1. Dependency and import path migration
2. Bubble Tea model contract changes
3. Bubble Tea and Bubbles behavior changes in shared base views
4. Huh form migration to the v2 API and v2 Lip Gloss types
- Fixes #757

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
- 
- 
-


qcserestipy:
@NishchayRajput Thank you for taking this on. The mechanical migration looks consistent, and moving alt screen into the view state is exactly how v2 intends it.

Testing against the demo Harbor instance surfaced two regressions, both in shared base views (so two fixes cover all call sites). Both have the same root cause: v2 clips to the terminal cell buffer, while v1 flushed the final frame regardless of size.

1. **`tablegrid` renders blank** (`robot create` permission grid): `tablegrid.New` never sets `table.WithWidth(...)`. In bubbles v2 the table's viewport defaults to width 0 and renders `""`. Fix: compute the width from the columns and pass `table.WithWidth(tableWidth)`, same as `tablelist/model.go` already does.

2. **`tablelist` cuts tall tables and leaks escape codes** (`project list` etc., 44 files): these views quit in `Init()`, so v2 clips the frame to terminal height, and the startup terminal queries (DECRQM) land in the shell as garbage (`2026;2$y...`). Fix: skip `tea.NewProgram` for these print-once views and use the pattern this PR already applies in `info.go`:

    ```go
    fmt.Println(m.View().Content)
    ```

For verification, one command per base class covers everything: `project list`, `artifact list`, `robot create`, `login`. `selection` already looks correct.

Follow-up material, fine to defer: `pkg/views/base/multiselect` is now dead code, and `tablegrid` could scroll instead of clip via `WindowSizeMsg`.

PR #1099, opened 2026-09-14
fix(webhook): avoid rendering interactive table when output format is…

### LINKED ISSUE #1098: [bug]: harbor webhook list cmd renders both interactive ui and json when format specified is json
## Description
When running `harbor webhook list [PROJECT_NAME]` with the `--output-format` flag (e.g., `-o json`), the command outputs the requested formatted JSON and then immediately proceeds to render the interactive Bubble Tea table UI directly below it.

## Steps to Reproduce
1. Run command: `harbor webhook list <project_name> --output-format json`
2. With arguments/config: valid project name with at least one webhook policy configured
3. Observe error: raw JSON is output, followed immediately by the interactive table view underneath it

## Expected Behavior
When `--output-format` (e.g. `json`, `yaml`, `csv`) is specified, only the requested formatted data stream should be printed to stdout.

## Actual Behavior
The formatted JSON is printed to stdout, followed immediately by the interactive Bubble Tea table UI rendering underneath it.

## Environment
- OS: Linux / All
- Tool version: current `main`
- Other relevant details: tested on local Harbor v2.15.2 instance

## Additional Context
In `cmd/harbor/root/webhook/list.go`, `webhookViews.ListWebhooks(resp.Payload)` is currently called unconditionally outside of the `output-format` check:

```go
FormatFlag := viper.GetString("output-format")
if FormatFlag != "" {
    err = utils.PrintFormat(resp, FormatFlag)
    if err != nil {
        return fmt.Errorf("failed to print in %s format: %v", FormatFlag, err)
    }
}
webhookViews.ListWebhooks(resp.Payload)
return nil

TaichKarna:
## Description
This PR fixes a bug where `harbor webhook list [PROJECT_NAME]` rendered the interactive Bubble Tea table UI in addition to formatted output when `--output-format` (e.g. `-o json`) was specified.

- Fixes #1098

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update
- [ ] Chore / maintenance

## Changes
- Placed `webhookViews.ListWebhooks` inside an `else` branch in `cmd/harbor/root/webhook/list.go` when `output-format` is specified.
- Grouped `listWebhooks` and `listWebhooksView` package variables in `list.go` to support unit testing.
- Added unit test for `webhook list` to verify `--output-format` behavior.
- Aligned `webhook list` behavior with other CLI list commands (`schedule list`, `robot list`, `registry list`, `repository list`, `project list`).

## Testing
- Verified locally with `./bin/harbor-cli webhook list demo -o json` against local Harbor instance. Only formatted JSON is printed.
- Added unit test covering `webhook list --output-format`.
- Formatted with `gofmt -s -w .` and verified full unit test suite passes: `go test ./...`.


JituRewar (commented):
@TaichKarna One suggestion from my side :  could we add a unit test covering - output-format (e.g. json) to verify that webhookViews.ListWebhooks is not invoked/rendered when a format is specified? This would help prevent the same regression from being introduced again in the future.
And the control-flow change looks good to me.

TaichKarna:
> @TaichKarna One suggestion from my side : could we add a unit test covering - output-format (e.g. json) to verify that webhookViews.ListWebhooks is not invoked/rendered when a format is specified? This would help prevent the same regression from being introduced again in the future. And the control-flow change looks good to me.

Yes, added one

